PR #13, opened 2014-12-30
Adds branch flag for changing which branch to download

russweas:
If you use the flag --b you can use whatever branch of laravel you like.

i.e. `laravel new mysite --b=develop`

I tried downloading it directly from github, but it said that it could not confirm the SSL certificate, even though I was using http:// ... :/ . So I used GH pages and it works fine. Hopefully if this gets passed Taylor puts it on laravel cabinet like the other link.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -105,7 +105,7 @@ class NewCommand extends \Symfony\Component\Console\Command\Command {
 	}
 
 	/**
-	 * Determine which file to download, 4 or 5
+	 * Determine which branch to download.
 	 *
 	 * @return string
 	 */


rweas:
OK so I was thinking about it and since the chances of getting into laravel cabinet are pretty low at this point, I'm going to upload my fork to packagist and host the code on my own personal site and have a cron job upload it nightly.

Before I do so I just want to check with @GrahamCampbell to make sure that this is an ok thing to do.
And @jallits I will give you instructions on how to install as soon as I'm done.

jallits:
Anyone think that maybe @taylorotwell might delegating some authority to another individual on this one? I think @rweas solution is great for the community. It scales with future releases and is backwards compatible.

rweas:
OK I've gotten a new lead on how to do this without cabinet. When running `curl https://github.com/laravel/laravel/archive/develop.zip` in iTerm, it says:
```
vagrant@homestead:~$ curl https://github.com/laravel/laravel/archive/develop.zip
<html><body>You are being <a href="https://codeload.github.com/laravel/laravel/zip/develop">redirected</a>.</body></html>
```
What this means is that I need to run `curl https://codeload.github.com/laravel/laravel/zip/develop` instead, which works perfectly.

rweas:
OK it's good now. It should be working perfectly for any branch! If no one finds any bugs or anything ( @jallits  and @GrahamCampbell ) I'll squash the commits and open a new PR!

jallits:
@rweas Problem is you cannot guarantee that curl or wget is available on every system (i.e. Windows)

rweas:
That's how laravel installer worked before though...

rweas:
OK final call before I squash it!

src/NewCommand.php
```diff
@@ -85,6 +105,19 @@ protected function download($zipFile)
 	}
 
 	/**
+	 * Determine which file to download, 4 or 5
+	 *
+	 * @return string
+	 */
+
```
GrahamCampbell: ?

src/NewCommand.php
```diff
@@ -85,6 +105,19 @@ protected function download($zipFile)
 	}
 
 	/**
+	 * Determine which file to download, 4 or 5
```
GrahamCampbell: missing full stop

rweas:
OK those are old comments I'll fix it to say which branch thanks. What does full stop mean though?

rweas:
I'm guessing it means a period.

PR #17, opened 2015-03-03
Add `--fresh` option to remove scaffolding

davidhemphill:
I'm not a fan of the default scaffolding, but recognize other people are, and I don't hold it against them, so I'd rather pass in a `--fresh` flag when creating a new project. That's just me. No judgement here. Actually, the auth scaffolding is pretty handy, I just don't prefer Bootstrap. Different strokes.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -3,6 +3,7 @@
 use ZipArchive;
 use Symfony\Component\Process\Process;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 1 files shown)


laravel
```diff
@@ -4,7 +4,7 @@
 require __DIR__.'/../../autoload.php';
 //require __DIR__.'/vendor/autoload.php';
 
-$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.0');
+$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.1');
```
GrahamCampbell: Bumping the version should not be done in pulls.

GrahamCampbell: Taylor will do this when he wants to do the next release.

davidhemphill: Gotcha, I'll remove that and force it up

taylorotwell:
Gonna hold on this until it becomes clear what the included assets stuff will look like in 5.1.

qw3rtman:
:+1: on this idea. Would be a great addition, I know I would use it.

PR #20, opened 2015-03-14
Guzzle 5 And Other Cool Stuff

GrahamCampbell:


taylorotwell:
Do you happen to know if Guzzle 4 and Guzzle 5 API's are equivalent for our use case. If we upgrade to 5 here it won't allow any other global CLI libs to use 4. So if possible we might want to consider allowing both 4 and 5.

GrahamCampbell:
Yeh, we could safely allow guzzle 4 here too.

PR #22, opened 2015-03-21
Proposed CS Fixes

GrahamCampbell:


GrahamCampbell:
k, not a fan then, lol

PR #24, opened 2015-04-09
Allowing crafting in an empty dir

phanan:
Right now the script doesn't allow crafting a new Laravel application into an empty directory – it throws an exception if the dir exists. I've added a check to enable this. 

With this change, these commands (which I tend to use) will work fine:

```
mkdir blog
cd blog
laravel new .
```


phanan:
Would you care to elaborate a bit on why, @taylorotwell?

PR #26, opened 2015-04-29
Keep symbolic links in zip

jildertmiedema:
The files in ```vendor/bin``` or symbolic linked.

Use the -y  for ```zip``` will keep the symbolic links.

Otherwise this will produce an error:
```sh
php laravel new test-app
cd test-app/vendor/bin
php ./phpunit
```

### DIFF
--- a/zipper.sh
+++ b/zipper.sh
@@ -2,7 +2,7 @@ wget https://github.com/laravel/laravel/archive/master.zip
 unzip master.zip -d working
 cd working/laravel-master
 composer install
-zip -r ../../laravel-craft.zip .
+zip -ry ../../laravel-craft.zip .
 cd ../..
 mv laravel-craft.zip public/laravel-craft.zip
 rm -rf working


taylorotwell:
This doesn't fix the problem.

jildertmiedema:
Maybe the ```zip``` command is distro dependent. I didn't see the symlinks in my gui zip viewer when looking at the generated zip, but after extracting the archive all was as expected.

jildertmiedema:
More people having problems.

Maybe we can do a ```rm vendor/bin/*``` in the ```zipper.sh``` script before the actual zip is generated. This will exclude the not working files/symlinks. So people won't be able to start them and don't see confusing errors.

On windows systems the bin folder looks different because the scripts will be created their as ```.bat``` files. 

Therefor everybody has to run ```php composer update``` if they want to use the vendor scripts.

syokunin:
What is the status of this issue?

taylorotwell:
What do we need to add to command. The last PR broke the build.

jildertmiedema:
by using `zip -ry ../../laravel-craft.zip` the symblic links or saved in the zip file.
Unziping this file works from cli `unzip latest.zip`
But using php `ZipArchive` won't keep the symblic links.

And windows does not even use these file file, composer creates a bat file when running `php composer install`

crynobone:
@taylorotwell would running `composer install` regenerate the bin folder? SInce `composer.lock` is available this process could actually be fast.

@Seldaek is there a way to regenerate broken symlink via composer command? Or can `composer install` rebuild/reset it if the vendor folder already exist.

Seldaek:
Just running install doesn't regenerate those (yet), only installing a package fresh would do that. As for symlinks yeah they are pretty bad for cross-platform. Maybe we should just generate proxy files even on linux so that they work everywhere.

HiroKws:
Only Japanese resource I found...

for symbolic links :
http://ja.stackoverflow.com/questions/11999/

for execute files:
http://ja.stackoverflow.com/questions/12037

Maybe, this will help the situations. I never met this problem, so I can't do anything about this.

PR #27, opened 2015-06-09
Call composer command to copy .env file

jaden:
If you run the installer to create a new project it says the
Application key was set successfully even though it wasn't.

This calls the composer script to copy .env.example to .env, so the
next script that calls artisan key:generate will work.

The reason for the current behavior is that php artisan key:generate
doesn't display an error if it fails to find a .env file.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -47,6 +47,7 @@ class NewCommand extends Command
 
         $commands = [
             $composer.' run-script post-install-cmd',
+            $composer.' run-script post-root-package-install',
             $composer.' run-script post-create-project-cmd',
         ];
 
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -46,8 +46,8 @@ class NewCommand extends Command
         $composer = $this->findComposer();
 
         $commands = [
-            $composer.' run-script post-install-cmd',
             $composer.' run-script post-root-package-install',
+            $composer.' run-script post-install-cmd',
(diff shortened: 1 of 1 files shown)


syokunin:
Command execution order is wrong.
'post-root-package-install' must be executed first.

https://github.com/laravel/laravel/commit/2ac993f8d369c0ca367accc93714ad960995f91e

````
$commands = [
+   $composer.' run-script post-root-package-install',
    $composer.' run-script post-install-cmd',
    $composer.' run-script post-create-project-cmd',
];
````

jaden:
Why do you say that? post-install-cmd runs "php artisan clear-compiled" and "php artisan optimize". Your order will work too, but it makes more sense to put the two dependent lines next to each other. 

The only requirement is that post-root-package-install runs before post-create-project-cmd so .env is in place before "php artisan key:generate" runs.

syokunin:
If installer do not copy the .env file first,
'php artisan optimize' would generate a 'compiled.php'.

Take a look at this issue.
https://github.com/laravel/laravel/commit/2ac993f8d369c0ca367accc93714ad960995f91e

Since .env have a 'APP_DEBUG = true',
'php artisan optimize' does not generate 'compiled.php'.

jaden:
Thanks for the clarification, I will push a new commit.

blackfyre:
:+1:

socketblind:
:+1:

PR #30, opened 2015-07-03
Added option to install Homestead per project along with installer

ArjanSchouten:
With these changes you can install Homestead with the installer:
```
laravel new Blog --with-homestead
```
This will add homestead to the application as described in the [docs](http://laravel.com/docs/5.1/homestead#per-project-installation).

Tested on Ubuntu 14.04 with php 5.6 and hhvm.
Tested on Windows with php 5.6.

(lumen-installer and docs needs pull-request too after this but wait with that right now :) )

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
     },
     "require": {
         "guzzlehttp/guzzle": "~4.0|~5.0",
+        "illuminate/support": "~5.1",
         "symfony/console": "~2.3",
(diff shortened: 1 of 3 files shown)


taylorotwell:
Gonna hold off on this. Just more to maintain :) ... I think it's fine to just do "composer require laravel/homestead"

ArjanSchouten:
True. Thanks for explanation.

PR #32, opened 2015-07-06
keep symlink and permission

syokunin:
This is fix for https://github.com/laravel/framework/issues/9491.


GrahamCampbell:
Adding more dependencies isn't really good since this is a global install so now we have more stuff to conflict.

PR #35, opened 2015-10-20
Parameters removed because they are not in use

anteriovieira:


GrahamCampbell:
:+1:

leandrorosendo:
:+1:

taylorotwell:
They are in use. I'm setting the timeout to `null`. (infinite timeout).

anteriovieira:
True, I am agree with you.

PR #36, opened 2015-10-20
Revert "Parameters removed because they are not in use"

taylorotwell:
Reverts laravel/installer#35


taylorotwell:
They are in use. I'm setting the timeout to null (no timeout). reverted.

GrahamCampbell:
lol

PR #37, opened 2015-10-31
Check if the given folder are empty before to install.

rbadillap:
It's just an additional condition to check if given folder are empty or not.
If the given folder (that could be *"."*) is empty, *installer* can craft the Laravel application there.
With that you can run something like:
`$ laravel new .`

Also I'm considering that the RuntimeException should say: **"Folder is not empty"** instead of "Application already exists!". But is not part of this PR. 

Let me know your thoughts.


taylorotwell:
Going to hold off on this.

PR #38, opened 2015-11-13
Allow use if Guzzle in version >= v6.0

okaufmann:


GrahamCampbell:
Have you tested it actually works?

PR #39, opened 2015-11-13
Allow use of Guzzle 6.0

okaufmann:
I've tested it on my homestead and a windows 10 machine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     },
     "require": {
-        "guzzlehttp/guzzle": "~4.0|~5.0",
+        "guzzlehttp/guzzle": "~4.0|~5.0|~6.0",
         "symfony/console": "~2.3",
         "symfony/process": "~2.3"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -1,28 +1,30 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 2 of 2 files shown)


laravel
```diff
@@ -7,7 +7,7 @@ if (file_exists(__DIR__.'/../../autoload.php')) {
     require __DIR__.'/vendor/autoload.php';
 }
 
-$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.1');
+$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.2');
```
GrahamCampbell: don't do that please

GrahamCampbell:
Please squash to one commit.

laravel
okaufmann: Should I leave it as it is?

vinkla: Yes, @taylorotwell or @GrahamCampbell will handle the versions.

okaufmann: Ok, sorry. Is there a Contribute Guide anywhere ?

vinkla: No worries. I don't there are any guidelines for this tool. It is common that the maintainers handle both versions and changelogs. We don't know if they want to add more stuff before the next release. We also don't know if this is going to be a major update or not.

okaufmann: U are right, thanks for make things clear ;)

PR #40, opened 2015-12-02
Implement lts command for 5.1 LTS project.

HiroKws:
Ref: laravel/laravel#11002

Implement lts command for 5.1 LTS project.

```
$> diff -u src/NewCommand.php src/LtsCommand.php  
--- src/NewCommand.php  2015-12-03 07:11:59.601960231 +0900
+++ src/LtsCommand.php  2015-12-03 07:19:23.960293175 +0900
@@ -11,7 +11,7 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class NewCommand extends Command
+class LtsCommand extends Command
 {
     /**
      * Configure the command options.
@@ -21,8 +21,8 @@
     protected function configure()
     {
         $this
-            ->setName('new')
-            ->setDescription('Create a new Laravel application.')
+            ->setName('lts')
+            ->setDescription('Create a new Laravel 5.1 LTS application.')
             ->addArgument('name', InputArgument::REQUIRED);
     }
 
@@ -94,7 +94,7 @@
      */
     protected function download($zipFile)
     {
-        $response = (new Client)->get('http://cabinet.laravel.com/latest.zip');
+        $response = (new Client)->get('http://cabinet.laravel.com/latest-lts.zip');
 
         file_put_contents($zipFile, $response->getBody());
```

zipper.sh made Zip file. However generated file has different name with in NewCommand class. ( 'laravel-craft.zip' in zipper.sh, but 'latest.zip' in download method of NewCommnad. So I suspected @taylorotwell has another work flow to create latest zip file, or zipper.sh is old.

Anyway it will need to change work flow up upload/generate zip files to 'cabinet.laravel.com' before accepted this pull request.

TODO: After accepted this pull request, there is need to change 'installtion' page for version 5.1 in document.

### DIFF
--- a/laravel
+++ b/laravel
@@ -7,7 +7,8 @@ if (file_exists(__DIR__.'/../../autoload.php')) {
     require __DIR__.'/vendor/autoload.php';
 }
 
-$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.2');
+$app = new Symfony\Component\Console\Application('Laravel Installer', '1.2.3');
 $app->add(new Laravel\Installer\Console\NewCommand);
+$app->add(new Laravel\Installer\Console\LtsCommand);
 
 $app->run();
--- a/src/LtsCommand.php
+++ b/src/LtsCommand.php
@@ -1,0 +1,152 @@
+<?php
+
+namespace Laravel\Installer\Console;
+
+use ZipArchive;
+use RuntimeException;
+use GuzzleHttp\Client;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class LtsCommand extends Command
+{
+    /**
+     * Configure the command options.
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('lts')
+            ->setDescription('Create a new Laravel 5.1 LTS application.')
+            ->addArgument('name', InputArgument::REQUIRED);
+    }
+
+    /**
+     * Execute the command.
+     *
+     * @param  InputInterface  $input
+     * @param  OutputInterface  $output
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->verifyApplicationDoesntExist(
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Don't duplicate the entire command please. Just add an option to the other command to choose the LTS version.

HiroKws:
Aha, there was uncertain conversation. 

My first opinion is using option(--lts), however Taylor mentioned 'lts acceptable' in larachat...

Maybe double meaning there. 'lts' as command? or 'lts' as 'option'??  Needed confirmation by @taylorotwell.

GrahamCampbell:
Even then, there's no way the command should be copy/pasted like this.

HiroKws:
Keep cool Graham.

If you don't have 'implement first, then refacter it' type idea, don't push your way. 

Now your are a 'guardian' of Laravel ( it means your a leader of Laravel community), your mention had strong influence to community. If somthing easy done by copy/paste, you must accept them. If not, community will lose intention to 'commit' something do easy.

And you said 'no way', so I read http://laravel.com/docs/5.1/contributions again. It never say like you pointed out. You must not added extra opinion with this. If @taylorotwell agreed it, please add the sentence to official document.

GrahamCampbell:
I don't really agree. Coping and pasting code is never acceptable in any context.

HiroKws:
OK, in honesty, once I had made base class and extends... but I found, this is just command. This is not library code. Also not framework code. Someone use this code? No. So, there is necessity to make them more complex now? No... There are possible to increase more download command? I don't think so...

If someone use this code, @taylorotwell might make more extendable. But this is just command. Keep it simple principal is the best principal. He did right work.

If this is a code of Illumination components, I totally agree with you. However this is just download command that it is few possibility to add more new commands.

Using base class has both of advantage/disadvantage. I think @taylorotwell feel need something big change for installer, he will refactor them by himself. 

(And I never say this loudly, sometime @taylorotwell copied some lines and forgot to correct comments. Anti-copy principal will attack him also... :) Criticism? No! I didn't start to fire.)

@taylorotwell : peaceful solution is that you reject this pull request, and implement by your self. It will spend your time a bit, however it bring big benefit to Laravel users. Specially me. In a few days, new Japanese Laravel book will start to print. So remaining just a few day to correct contents. When you make decide which is the better, 'laravel new --lts' or 'laravel lts', let's me know. Please.

taylorotwell:
I think probably we would just want a flag.

PR #45, opened 2016-02-07
Throw an exception instead of a fatal error for the Zip extension

saulotoledo:
Before this fix, creating a new application throws a fatal error if the Zip extension is not installed.


saulotoledo:
Invalid. See #46 .

PR #47, opened 2016-03-25
Allow TTY output from process

alexbowers:
This allows us to get the coloured output from Process.

PR #48, opened 2016-03-27
Bug fixes for windows machines

alexbowers:


taylorotwell:
What happens on Windows machines without these checks?

m1r0:
@taylorotwell see <https://github.com/laravel/installer/pull/47#issuecomment-201965753>

I have the same issue.

GrahamCampbell:
This change is incorrect.

GrahamCampbell:
Some Windows systems do support this.

alexbowers:
This chance is coming directly from what symfony do to check and throw the exceptions. Not sure how else we can fix this.

alexbowers:
https://github.com/symfony/process/blob/master/Process.php#L913

taylorotwell:
So if they are doing the check why do we also need to do the check?

arrilot:
We can simply wrap the call in try-catch instead of copypasting all those crappy checks from symfony

RobWJacob:
this patch works on my windows 10 setup. thank you for the fast fix ;)

alexbowers:
@arrilot Im not a big fan of a try catch, because in the future if there was a different reason for an exception to be thrown, we wouldn't necessarily know about it.

alexbowers:
@taylorotwell Our check is the opposite of theirs, because we only want it to happen if tty is allowed, whereas process throw exceptions. We either need to only run the function if tty is enabled (the if), or catch the exception.

taylorotwell:
Can someone answer the question above? They are already donig the checks, so why do we have to?

arrilot:
@taylorotwell As far as I can see they explicitly  throw runtime exceptions if tty cant be set. As a result laravel installer may die with these exceptions atm because we neither prevent them nor catch them

alexbowers:
@taylorotwell If you notice, the conditions im checking for are the exact opposite of what Symfony checks for to determine support.

Since we never want it to throw the exception (we just want to disable the feature instead), I am checking for whether any exception would be thrown using the same (but opposite) conditions of Symfony so that we don't even attempt TTY support if it would result in an exception being thrown.

alexbowers:
If we don't do this, then as @m1r0 and @RobWJacob (on https://github.com/laravel/installer/pull/47#issuecomment-201965753) said, it stops the installer working for them on Windows because it throws an exception saying that TTY isn't supported on Windows. 

Instead of throwing the exception, we just want it to fallback to non-tty.

svenluijten:
I'm still having this issue, can this be tagged so we can pull this in with a simple `composer global update`?

PR #49, opened 2016-03-30
Allow install on empty directories

alexbowers:
There have been a fair few times where I have created my directory before hand (eg. Shared / Synced folder in vagrant), where it is tricky to `rm -rf $directory`, but the directory is completely empty. Currently, because it is a directory, it throws an exception, but install should be fine if the folder is clean.


taylorotwell:
Going to hold off on this.

PR #50, opened 2016-03-30
Throw error if a file exists with that name

alexbowers:
Currently there is only checks for if a directory exists with that name, however, we don't check whether a file exists with the name. If it does, then you get a ZipArchive error:

image


alexbowers:
Didn't know what to do about wording, so that is up for input.

PR #53, opened 2016-07-27
Fix composer.par path.

hikouki:


vlakoff:
Maybe the path should be enclosed in double quotes?

hikouki:
@vlakoff 

Thank you for replay :)

Fix it.

GrahamCampbell:
Does this work on all operating systems?

vlakoff:
Mmmm, at least on Windows, it seems to work as well without this added `getcwd()`.
Also, you shouldn't have removed the double quotes around PHP_BINARY.

Unless a failing example is provided, looks like the current code is just fine.

hikouki:
Thank you for test 🍺 

Added the double quotes around PHP_BINARY.

taylorotwell:
The current code seems fine unless someone has a serious bug with it.

hikouki:
@taylorotwell 

Isn't this bug?

for MacOSX

PR #58, opened 2016-12-09
Fix PHP docs

jerguslejko:
This just fixes the phpdocs and type hinting for `getVersion()` method.

🌵

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -32,8 +32,8 @@ class NewCommand extends Command
     /**
      * Execute the command.
      *
-     * @param  InputInterface  $input
-     * @param  OutputInterface  $output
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -171,10 +171,10 @@ protected function cleanUp($zipFile)
     /**
      * Get the version that should be downloaded.
      *
-     * @param  \Symfony\Component\Console\Input\InputInterface  $input
+     * @param  InputInterface  $input
```
GrahamCampbell: please revert

jerguslejko: any particular reason? `execute(...)` method uses short names as well

GrahamCampbell: Using short names was a mistake. We always use fully qualified class names.

PR #59, opened 2017-01-20
Artisan command

litvinchuk:
What do you think about this small feature?

This update allows run artisan commands from any directory of project.

**For example**:
```
cd /path/to/project/app/Http/Controllers;
laravel artisan make:controller SomeController;
```


taylorotwell:
Not planned at this time.

PR #61, opened 2017-01-27
Switch installer to use GitHub zips

simonhamp:
This all started with: https://larachat.slack.com/archives/laravel5/p1485514747011330

Then I found the `--5.2` option didn't work at all anymore.

Decided to switch to GitHub's zips and simplified the command.

Result: can now install any defined branch from GitHub using the installer by using the new `--ver` command, e.g.:

    laravel new my-project-name --ver=5.3

Hope this is acceptable - I haven't checked whether it's allowed etc on GitHub's TOS 😅

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -122,7 +122,7 @@ class NewCommand extends Command
                 $filename = 'latest.zip';
                 break;
             default:
-                $filename = 'latest-' . $version . '.zip';
+                $filename = 'latest-'.str_rpelace('.', '', $version).'.zip';
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The main problem is this sort of ruins the whole premise of the installer which was to have the composer.lock file in the zips (which the cabinet server builds) and it makes the installation of dependencies super fast. :)

simonhamp:
Ah! Boo... I wasn't aware of that.

I understand the point, but it loses some flexibility if those builds aren't done on your server, for example, the 5.2 option is broken right now. Plus with 5.1 being LTS, it would be nice to be able to install that via CLI.

How are you packaging the builds on the cabinet server? Is that automated or manual?

taylorotwell:
It's automated hourly. I just need to add the other versions to the job.

PR #63, opened 2017-04-11
Add --force option to ignore existing directory in new command

jodiedunlop:
When you already have an existing directory under source control and want to install laravel to it, it can be a little bit tedious since the installer wont write to an existing directory.

To handle this use case, a --force option is provided which ignores if the directory already exists and proceeds to install to that directory.


rtconner:
Can we get a new version of the installer, with this feature added? I would like to use it.

PR #65, opened 2017-06-14
Add an option to get latest lts release

m1guelpf:
Adds an option to get the latest LTS release.

**NOTE: This would need a lts.zip file on cabinet.laravel.com**

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -190,7 +190,7 @@ class NewCommand extends Command
         }
         
         if ($input->getOption('lts')) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to add this right now.

lk77:
I think it would be important to have multiple versions like lts.
In this moment latest.zip causes problems because there is a composer.lock inside.
Provide a lts option, or better, provide an option to clone from github should be done i think, to provide a fallback when latest release fail.

PR #67, opened 2017-07-08
Set permissions for directories that need to be writable by server

peterojo:
I find my self having to manually set the permissions on the `storage` and `bootstrap/cache` folders after each new install. Would be nice if it the installer at least attempts to do this automatically and gives a nice reminder if it can't. My first Laravel PR
Huge fan.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "ext-zip": "*",
         "guzzlehttp/guzzle": "~4.0|~5.0|~6.0",
         "symfony/console": "~2.3|~3.0",
-        "symfony/process": "~2.3|~3.0"
+        "symfony/process": "~2.3|~3.0",
+        "symfony/filesystem": "^3.3"
     },
     "bin": [
         "laravel"
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
         "ext-zip": "*",
         "guzzlehttp/guzzle": "~4.0|~5.0|~6.0",
         "symfony/console": "~2.3|~3.0",
-        "symfony/process": "~2.3|~3.0",
-        "symfony/filesystem": "^3.3"
+        "symfony/filesystem": "~2.3|~3.0",
+        "symfony/process": "~2.3|~3.0"
     },
     "bin": [
         "laravel"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -18,7 +18,8 @@
         "ext-zip": "*",
         "guzzlehttp/guzzle": "~4.0|~5.0|~6.0",
         "symfony/console": "~2.3|~3.0",
-        "symfony/process": "~2.3|~3.0"
+        "symfony/process": "~2.3|~3.0",
+        "symfony/filesystem": "^3.3"
```
GrahamCampbell: Should be `"~2.3|~3.0"` please.

GrahamCampbell: Also, please put this in between console and process, so it's in the correct order.

src/NewCommand.php
```diff
@@ -6,11 +6,14 @@
...
 use GuzzleHttp\Client;
 use Symfony\Component\Process\Process;
+use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+
```
GrahamCampbell: One too many new lines.

src/NewCommand.php
```diff
@@ -174,6 +178,26 @@ protected function cleanUp($zipFile)
     }
 
     /**
+     * Make sure the storage and bootstrap cache directories are writable
```
GrahamCampbell: Missing fullstop and blank comment line.

GrahamCampbell:
Thanks for the PR. Left some comments.

composer.json
peterojo: Thanks. I'll make these changes and PR again

taylorotwell:
Do you really need them to be set to 777?

peterojo:
I just make them 777 for development, but I guess 775 would be more appropriate, or what do you suggest?

taylorotwell:
You should probably use `Illuminate\Filesystem\Filesystem`.

peterojo:
But @taylorotwell the `Illuminate\Filesystem\Filesystem`'s `chmod()` method doesn't allow me to do it recursively, and it also doesn't allow me to catch the exception if it fails.

PR #69, opened 2017-07-27
Create composer.lock as minimun required platform

gmsantos:
Create composer.lock emulating minimum PHP version for Laravel 5.4 avoid errors like the reported here https://github.com/laravel/framework/issues/20214 , keeps the lock file for faster installations and allow to [relax dependencies constraint again](https://github.com/laravel/framework/pull/20259).

@taylorotwell is this `.sh` still used to generate http://cabinet.laravel.com/ files? It's not used on NewCommand as far I know.

Related:
https://getcomposer.org/doc/03-cli.md#config
https://getcomposer.org/doc/06-config.md#platform
https://github.com/laravel/framework/issues/20214
https://github.com/laravel/framework/pull/20241#issuecomment-317564201
https://github.com/laravel/installer/pull/68

### DIFF
--- a/zipper.sh
+++ b/zipper.sh
@@ -2,6 +2,7 @@
 wget https://github.com/laravel/laravel/archive/master.zip
 unzip master.zip -d working
 cd working/laravel-master
+composer config platform.php 5.6.4
 composer install
 zip -ry ../../laravel-craft.zip .
 cd ../..


lk77:
i don't think it's the solution, because it force users of php ~7.0 to do a composer update to have most recent packages. i think we need a php5.x-latest.zip and a php7.x-latest.zip

gmsantos:
They will not get the latest only for packages that require php >= 7.0. All the other packages will came with the most recent version defined by the constraint.

Having a cabinet package by PHP version will require minor versions too (php7.1-latest.zip, php7.2-latest.zip), even patch because this requirement is in package domain. It's a nice solution but depends to update the package builder, modify the `NewCommand` and release a new tag for Laravel Installer.

With this approach we only have to include a new line on the `.sh` package builder. The drawback is to do a composer update on newer PHP versions (PHP 7.0+) vs breaking a supported version (PHP 5.6 with Laravel 5.4). At least all scenarios will have a up and running Laravel project.

taylorotwell:
I'm not sure this is needed anymore since we're requiring a more specific version of the inflector now?

lk77:
yes it's actually fixed, but the problem remain, if another package drop the support of php5.6, it will happens again.

themsaid:
@taylorotwell yes I believe this isn't needed anymore

taylorotwell:
Added this to the laravel build server.

gmsantos:
@taylorotwell @themsaid can https://github.com/laravel/framework/pull/20227 be reverted?

Also if `zipper.sh` isn't used anymore I suggest to remove it from installer.

PR #73, opened 2017-12-09
Updated findComposer function to work with composer.phar installed at home dir

clytras:
Changed findComposer function to check if there is a composer.phar file inside the HOME dir "~/composer.phar". Some CPanel installations keep older composer version when updating and laravel installer uses the old versions to run through. Check laravel/framework#20864 (comment) for more information. 

Previous pull request closed because of a mistype that had the `file_exists` function to just check the `$_SERVER['HOME']` without concating the `'/composer.phar'` inside the function after the home dir.


clytras:
I thought this was standard for CPanel, to install `composer.phar` to the HOME directory, which I realize now it's not the case. Closing.

PR #74, opened 2017-12-11
Remove duplicated command

mbabker:
Unless I'm missing something obvious there should be no need to run `post-autoload-dump` a second time for dev.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -70,8 +70,6 @@ class NewCommand extends Command
 
         if ($input->getOption('dev')) {
             unset($commands[2]);
-
-            $commands[] = $composer.' run-script post-autoload-dump';
         }
 
         if ($input->getOption('no-ansi')) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
@themsaid any idea why this might have been here or was it a mistake?

themsaid:
I think the entire block should be removed:

```
if ($input->getOption('dev')) {
    unset($commands[2]);

    $commands[] = $composer.' run-script post-autoload-dump';
}
```

All these commands has to run in master and dev:

```
$composer.' install --no-scripts',
$composer.' run-script post-root-package-install',
$composer.' run-script post-create-project-cmd',
$composer.' run-script post-autoload-dump',
```

mbabker:
It originally came from https://github.com/laravel/installer/pull/64 so based on the comments there the `post-create-project-cmd` script wouldn't be needed at all now.

themsaid:
@mbabker but that script creates the app key.

mbabker:
Ah, ya, just had to look to double check.  If block is gone in full now.

PR #76, opened 2018-02-14
Add --lts command option to install 5.5 LTS version

nafiesl:
I see there are 2 PR here which proposed lts command option. The PR requires to have new lts.zip file on the cabinet.

In this PR, the --lts command will get zip file from latest 5.5 branch. So it doesn't requires new files from the laravel cabinet.

shutter_20180214-012

Thanks for reviewing this PR.


taylorotwell:
No plans to add right now.

PR #78, opened 2018-10-13
Add new clone command

sarthaksavvy:
* This will clone the repository and install composer init.

* If the repository is a project then generate key for it.

Reduces 3 steps into 1.


taylorotwell:
No plans on adding this right now.

PR #83, opened 2019-03-11
Replace database name with the name of the app

pascalbaljet:
This PR updates the `DB_DATABASE` value in `.env` and `.env.example` to the given name of the app. This is one of the first things I do when I create a new Laravel project and I thought it would be great if the installer did it for me. I think most developers have more than one project installed locally, so `homestead` will almost always be replaced.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -81,7 +81,7 @@ class NewCommand extends Command
             }, $commands);
         }
         
-        $commands[] = "sed -i '' 's/DB_DATABASE=homestead/DB_DATABASE={$name}/g' .env";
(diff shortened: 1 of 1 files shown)


taylorotwell:
The `sed` command may not always be available?

pascalbaljet:
It's available on Unix systems (including macOS). On Windows it's included in Git BASH and the (official) Windows Subsystem for Linux. Must confess I haven't used Windows for a while so I don't know what people use these days.

taylorotwell:
is it not possible to just use a `str_replace` on the file after it's written?

taylorotwell:
Closing pending above changes.

PR #85, opened 2019-03-15
Update composer path on findComposer

peter279k:
# Changed log
- Update the composer execution path on `NewCommand::findComposer` command.

Actually, the original `php compoer.phar` string result is not correct.

And this will not specify the current path to let PHP binary know where the `composer.phar` can execute.

It should use the current `composer` path to tell PHP binary to execute `composer.phar`.

# Instruction

## Environment
- Ubuntu 18.04
- composer version is `1.8.4`.

Here are steps about reproducing the issue.

- Firstly, we use the `composer require laravel/installer` command.
- Then using the `vendor/bin/laravel new blog` to create the new Laravel project.
- It will get the error message in following image during building the Laravel project.

image

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -219,8 +219,10 @@ class NewCommand extends Command
      */
     protected function findComposer()
     {
-        if (file_exists(getcwd().'/composer.phar')) {
-            return '"'.PHP_BINARY.'" composer.phar';
+        $composerPath = getcwd().'/composer.phar';
+
+        if (file_exists($composerPath)) {
+            return '"'.PHP_BINARY.'" '.$composerPath;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please provide instructions for how to recreate the bug locally.

taylorotwell:
Closing pending inactivity. Feel free to re-open a PR with reproduction instructions.

peter279k:
I have already created the instructions on the original comments.

driesvints:
@peter279k you should have pinged Taylor about that. We can't know if you updated the original comment unless we check the "edited" history (which we don't always do). I see you've already created a new PR, thanks 👍

PR #87, opened 2019-03-18
Replace database name with the name of the app

pascalbaljet:
This PR replaces the previous PR (https://github.com/laravel/installer/pull/83). As suggested I've updated the script the use `str_replace` instead of `sed`. Here's the description of the original PR:

> This PR updates the DB_DATABASE value in .env and .env.example to the given name of the app. This is one of the first things I do when I create a new Laravel project and I thought it would be great if the installer did it for me. I think most developers have more than one project installed locally, so homestead will almost always be replaced.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -43,8 +43,9 @@ class NewCommand extends Command
         if (! extension_loaded('zip')) {
             throw new RuntimeException('The Zip PHP extension is not installed. Please install it and try again.');
         }
-
-        $directory = ($input->getArgument('name')) ? getcwd().'/'.$input->getArgument('name') : getcwd();
+        
+        $name = $input->getArgument('name');
+        $directory = $name ? getcwd().'/'.$name : getcwd();
 
         if (! $input->getOption('force')) {
             $this->verifyApplicationDoesntExist($directory);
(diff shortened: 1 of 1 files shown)


peter279k:
The commits are very large.

Can you merge all commits to one?

And it seems that these commits are not good.

You should check that.

pascalbaljet:
> Can you merge all commits to one?

Done

src/NewCommand.php
```diff
@@ -89,6 +90,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $process->run(function ($type, $line) use ($output) {
             $output->write($line);
         });
+        
+        $this->replaceDefaultDatabase("{$directory}/.env", $name);
+        $this->replaceDefaultDatabase("{$directory}/.env.example", $name);
```
peter279k: Is it necessary to replace default database on `.env.example` file?

The `.env.example` is about the sample `.env` file and I think it doesn't have to replace the default with current database name.

pascalbaljet: The whole idea of this PR is to save time. The `.env` file is [ignored](https://github.com/laravel/laravel/blob/master/.gitignore#L6) by default, so every time someone clones the git repository of a project, they copy `.env.example` to `.env`. Not updating `.env.example` still requires that you manually update the database in `.env`.

peter279k: @pascalbaljet , thank you for your reply. That makes sense.

codeprofis: I don't think it makes sense to copy this information into .env.example. .env.example is not gitignored by default. This would increase the risk of users pushing internal configuration into git repositories.

It wouldn't be an insane risk because it's "only" the database name, but it would still be unnecessary.

taylorotwell:
I think I'm just going to hold off on this for now.

PR #89, opened 2019-04-19
[2.0] Update NewCommand

browner12:
I'm weirdly excited about this PR, hope you guys like it, too.

### Problem

Currently, we are maintaining a zip archive of the 'master' and 'develop' branches of the `laravel/laravel` project to be used when installing a new Laravel application through the Laravel Installer at the following URLs:

http://cabinet.laravel.com/latest.zip
http://cabinet.laravel.com/latest-develop.zip

I'm not positive, but updating these seems to be a manual process.  This makes it a little difficult for them to stay synced with the releases in Github. For example, if you go install a new Laravel app with the installer *right now*, you will end up with a `yarn.lock` file in your app, which no longer exists in the `laravel/laravel` repo, which means the Cabinet files are out-of-date.

Also, this setup limits us to installing only the current 'master' or 'develop' branches, not any prior branches, especially the LTS branch.

### Solution

Github actually maintains URLs for zip archives of projects. It seems like the solution is to simply switch the download URL from our current internally maintained one to the Github maintained one, but there is one small caveat.

Github wraps the files in an extra directory with the name of the project and the release. Kind of annoying, but not unfixable. PHPs `ZipArchive` class allows us to loop through the archive and remove this top directory. Great, now we can switch to Github's URL without making significant changes to the rest of the install process.

Because Github maintains Zip archives for **all** releases, now we can add an option to install a specific release, or use the new `lts` option to install the latest LTS release. I'm not sure why someone would want to install an old non-LTS version, but who knows.

### Considerations

- Technically the zip download URL for Github projects lives at `https://github.com/{org}/{repo}/archive/{release}.zip`. However, this URL 301s and ends up at the URL I used in the `download` method. Would we rather use the public facing URL, or continue to use the final URL and just update it if Github switches it?
- The `.gitattributes` file determines if any files are ignored in the export. Currently, we just ignore the `CHANGELOG.md`, which I think is good for now.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -28,6 +28,8 @@ class NewCommand extends Command
             ->setDescription('Create a new Laravel application')
             ->addArgument('name', InputArgument::OPTIONAL)
             ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
+            ->addOption('lts', null, InputOption::VALUE_NONE, 'Installs the latest "LTS" release')
+            ->addOption('release', null, InputOption::VALUE_OPTIONAL, 'Installs the requested release')
             ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists');
     }
 
@@ -52,9 +54,7 @@ class NewCommand extends Command
 
         $output->writeln('<info>Crafting application...</info>');
 
-        $version = $this->getVersion($input);
-
-        $this->download($zipFile = $this->makeFilename(), $version)
+        $this->download($zipFile = $this->makeFilename(), $this->getVersion($input))
              ->extract($zipFile, $directory)
              ->prepareWritableDirectories($directory, $output)
              ->cleanUp($zipFile);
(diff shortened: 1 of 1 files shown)


browner12:
ooph, looks like the StyleCI settings are messed up.

browner12:
I also see that this repo does not have a PHP version constraint in the `composer.json` file, and I used the null coalescing operator in here. If this gets merged, I will add a `^7.1.3` constraint.  If we add that constraint, maybe we need to do a major bump on this repo.

GrahamCampbell:
Please rebase. :)

driesvints:
The actual reason for the cabinet zips is that they get pre-installed dependencies so this speeds up the installation process. Github installs don't have this and with these changes the installer actually doesn't provides complete installations as they do now and would be a lot slower if we migrated the dependency processing as well.

taylorotwell:
What @driesvints said basically. They aren't manually built. They are built every hour or so. The yarn.lock file is just because we are running yarn on the server so that you get a lock file for faster installation on your side. However, this was before NPM had lock files so we can probably remove that now and just use NPM.

browner12:
thanks for the clarification guys

browner12:
is the `zipper.sh` script what handles this, or is the script not open source? would be nice to tweak it so we could allow the installation of specific releases.

driesvints:
Don't think so. Seems outdated tbh.

browner12:
should we get rid of it to avoid confusion? was that the point of the zipper originally? or was it supposed to be for people who don't use Composer?

driesvints:
I've pinged @taylorotwell about it.

browner12:
looks like Lumen got rid of it a looonnng time ago.

https://github.com/laravel/lumen-installer/commit/7b153b6b2d19c1e647fc10238f9497d47855c392

PR #90, opened 2019-04-22
remove temporary variable

browner12:


peter279k:
It seems that coding style is failed on StyleCI.

It should be checked.

browner12:
yah, they've made some changes to StyleCI recently, and I hadn't rebased. the failures don't have to do with this code.

driesvints:
Feel free to resubmit when you've rebased.

PR #91, opened 2019-04-22
merge Lumen installer and Laravel installer

browner12:
The Laravel and Lumen installer packages have about 99% of their code shared. This PR merges them, with the intention that we deprecate and drop support for the Lumen installer repo.

Installing a Laravel application will remain the same:

```sh
laravel new MyApp
```

Lumen installation will change slightly. Instead of:

```sh
lumen new MyApp
```

it will become:

```sh
laravel new:lumen MyApp
```

Merging these 2 packages will make sure they are both getting maintained and updated, and will reduce the maintenance load by dropping the Lumen repo.

There is probably the ability to abstract a parent class out of these 2 commands now, but I wanted to start with this and see if you guys are on board.

We could also tweak the naming if you think `laravel new:lumen MyApp` is too verbose. One option would be `laravel lumen MyApp`

### DIFF
--- a/laravel
+++ b/laravel
@@ -8,6 +8,7 @@ if (file_exists(__DIR__.'/../../autoload.php')) {
 }
 
 $app = new Symfony\Component\Console\Application('Laravel Installer', '2.0.1');
-$app->add(new Laravel\Installer\Console\NewCommand);
+$app->add(new Laravel\Installer\Console\NewLaravelCommand);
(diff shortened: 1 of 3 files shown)


driesvints:
I know you mean well but I wouldn't do this atm. I'd still keep everything Lumen and Laravel separate for the time being.

browner12:
any particular reason?

driesvints:
Because I believe it's better to keep these two separate like we do for the framework & the skeleton.

PR #93, opened 2019-05-15
Idea to allow preset package installs using a JSON file at the user root

danmatthews:
This pull request begins the ability to allow users to specify a `laravel-installer.json` file in their home directory to facilitate quickly pulling in extra packages when spinning up a new project.

The following format, using the format `prefix => require` is suggested:

```json
{
    "media": {
        "spatie/laravel-medialibrary": "^7.0.0"
    }
}
```

When creating a new install using `laravel new`, you can use the `--preset` option to include the extra packages: `laravel new --preset=media`.

This would:
1) Allows users to quickly (and portably) create distributions that include packages they use frequently.
2) Allow them to share this between a team or other framework users using a single simple file or JSON snippet.
3) Save time when spinning up new projects.

Currently doesn't do `dev` dependencies, but we can add that in as part of the JSON configuration if needs be.

Just preliminary thoughts - i'd definitely use this tool myself, so looking for ideas and feedback.

Thanks.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -28,7 +28,8 @@ class NewCommand extends Command
             ->setDescription('Create a new Laravel application')
             ->addArgument('name', InputArgument::OPTIONAL)
             ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
-            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists');
+            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists')
+            ->addOption('preset', null, InputOption::VALUE_REQUIRED, 'Choose a preset list of packages to install along with your new Laravel installtion');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -28,7 +28,8 @@ protected function configure()
             ->setDescription('Create a new Laravel application')
             ->addArgument('name', InputArgument::OPTIONAL)
             ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
-            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists');
+            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists')
+            ->addOption('preset', null, InputOption::VALUE_REQUIRED, 'Choose a preset list of packages to install along with your new Laravel installtion');
```
Jamesking56: I don't think it should be required, as we may still want to install plain Laravel.

danmatthews: You can still run `laravel new`, but providing the option without a value is not allowed:

okay: `laravel new --preset=media`
not okay: `laravel new --preset`

src/NewCommand.php
```diff
@@ -40,10 +41,36 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+
         if (! extension_loaded('zip')) {
             throw new RuntimeException('The Zip PHP extension is not installed. Please install it and try again.');
         }
 
+        $presets_path = getenv('HOME').DIRECTORY_SEPARATOR."laravel-installer.json";
```
Jamesking56: `$HOME` is usually not set in Windows by default.

src/NewCommand.php
```diff
@@ -40,10 +41,36 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+
```
Jamesking56: Unnecessary gap

src/NewCommand.php
```diff
@@ -91,6 +147,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $output->writeln('<comment>Application ready! Build something amazing.</comment>');
     }
 
+
+    public function doesPresetFileExist()
```
Jamesking56: Is this method supposed to be completed?

src/NewCommand.php
danmatthews: I haven't coding standards'ed (it's a word!) anything yet, i will once the feedback is in on filenames / handling etc.

src/NewCommand.php
danmatthews: Cool, i'll look into this - might be a use case for the posix functions.

src/NewCommand.php
danmatthews: This is just some pre-abstraction before i made the decision to seek feedback first.

src/NewCommand.php
Jamesking56: Possibly this: https://stackoverflow.com/a/21473475/941446

taylorotwell:

PR #94, opened 2019-05-24
Verify directory is also empty

arjasco:
Running `laravel new` without specifying a directory will install the framework in the current working directory.

### What this change does:

Checks if the directory is empty before proceeding which i imagine 99% of the time is what you want.

### Benefits

* Prevent you from polluting another directory you didn't mean to as it already contains files.
* Prevents you from overriding an existing laravel application in the CWD.

Yes, version control is a wonderful thing for the second point, but no help if you haven't already initialised a repo ie. testing a quick concept

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -92,6 +92,17 @@ class NewCommand extends Command
     }
 
     /**
+     * Check that the directory is empty.
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to add this right now.

PR #97, opened 2019-09-18
[Feature] Add --auth option to 'new' command

ryangjchandler:
## Problem
Although there is no actual problem here, this feature was request on the laravel/ideas issue board (https://github.com/laravel/ideas/issues/1833). 

I think this is actually a nice and clean addition to the installer, proving how easy it is to get a full authentication system running in Laravel.

## Solution
I have added a new option to the command, `--auth`, which will install the laravel/ui package and run the `php artisan ui:auth` command.

As @mattstauffer mentioned on the issue, these extra steps can already be removed by using Lambo or a custom Bash alias, but including it with the installer itself is, in my opinion, cleaner and easier for the user.


taylorotwell:
I don't think I'm going to integrate these things. The UI command has many more options.

PR #103, opened 2019-11-23
Symfony 5 Support

vadimski:
Added Symfony 5 Support

#SymfonyHackday


driesvints:
Thanks! This however misses the return integer for the execute function. I'll send in a new PR in a bit with some other updates as well.

PR #108, opened 2019-12-21
require symfony/process 4.2

crishoj:
`NewCommand` relies on `Process::fromShellCommandline()`, which was introduced in `symfony/process` version [4.2.0](https://github.com/symfony/process/commit/d6f417d21efe2a2b91c0c8a19d938c5db18d81d3).


GrahamCampbell:
Thanks for catching this. 👍

PR #109, opened 2020-01-02
Add Laravel packages on creating a new project

regiszanandrea:
Hi,

This PR references to idea [#1987](https://github.com/laravel/ideas/issues/1987)

This PR add an option to laravel/installer new command that install Laravel's packages after project creation. 
Only Laravel Packages are allowed, excluding paid projects.

The option name is `with` but could be other.

An example:
`$ laravel new blog --with=ui,horizon,telescope,socialite`

This will install `ui`, `telescope` and `socialite` packages and run their `install` commands like: `php artisan telescope:install`.

The package `ui` will be installed in `require-dev` section of composer.json respecting the official documentation.

Also, I improved some code style on NewCommand Class.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -297,13 +297,13 @@ class NewCommand extends Command
         $composerPackages = '';
         foreach ($laravelPackages as $laravelPackage) {
             if ($dev) {
-                if (isset(self::LARAVEL_DEV_PACKAGES[$laravelPackage]) ) {
+                if (isset(self::LARAVEL_DEV_PACKAGES[$laravelPackage])) {
                     $composerPackages .= ' '.self::LARAVEL_DEV_PACKAGES[$laravelPackage];
                 }
                 continue;
             }
 
-            if (isset(self::LARAVEL_PACKAGES[$laravelPackage]) ) {
+            if (isset(self::LARAVEL_PACKAGES[$laravelPackage])) {
                 $composerPackages .= ' '.self::LARAVEL_PACKAGES[$laravelPackage];
             }
         }
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -39,7 +39,6 @@ class NewCommand extends Command
         'dusk' => 'dusk:install',
         'scout' => 'vendor:publish --provider="Laravel\Scout\ScoutServiceProvider"',
     ];
-    
     /**
      * Configure the command options.
      *
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -106,18 +106,18 @@ class NewCommand extends Command
             }, $commands);
         }
 
-        if ($input->getOption('with')) {
-            $laravelPackages = $this->getLaravelPackages($input->getOption('with'), $composer);
+        if ($with = $input->getOption('with')) {
+            $laravelPackages = $this->getLaravelPackages($with, $composer);
             if ($laravelPackages) {
                 $commands[] = $laravelPackages;
             }
 
-            $laravelDevPackages = $this->getLaravelDevPackages($input->getOption('with'), $composer);
+            $laravelDevPackages = $this->getLaravelPackages($with, $composer, true);
             if ($laravelDevPackages) {
                 $commands[] = $laravelDevPackages;
             }
             if ($laravelPackages || $laravelDevPackages) {
-                $commands[] = $this->getLaravelInstallCommands($input->getOption('with'));
+                $commands[] = $this->getLaravelInstallCommands($with);
             }
         }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Please fix the phpdoc, in particular, to match Laravel's standards.

Don't worry about the rest of the code style though. It gets auto-fixed on merge. Easier for everyone that way. Contributors don't need to worry, and neither do maintainers.

src/NewCommand.php
```diff
@@ -243,4 +284,73 @@ protected function findComposer()
 
         return 'composer';
     }
+
+    /**
+     * @param string $packages
+     * @param string $composer
+     * @return string
+     */
+    private function getLaravelPackages(string $packages, string $composer)
```
GrahamCampbell: Please remove the string type here

src/NewCommand.php
```diff
@@ -243,4 +284,73 @@ protected function findComposer()
...
+
+        return $composerPackages;
+    }
+
+    /**
+     * @param string $packages
+     * @param string $composer
+     * @return string
+     */
+    private function getLaravelDevPackages(string $packages, string $composer)
```
GrahamCampbell: And these

src/NewCommand.php
```diff
@@ -243,4 +284,73 @@ protected function findComposer()
...
+        }
+
+        return $composerPackages;
+    }
+
+    /**
+     * @param string $packages
+     * @return string
+     */
+    private function getLaravelInstallCommands(string $packages)
```
GrahamCampbell: and this one

src/NewCommand.php
```diff
@@ -243,4 +284,73 @@ protected function findComposer()
...
+
+        if ($composerPackages !== '') {
+            return $composer . ' require --dev' . $composerPackages;
+        }
+
+        return $composerPackages;
+    }
+
+    /**
+     * @param string $packages
```
GrahamCampbell: Please put 2 spaces before and after the word string in param docs, in all cases.

regiszanandrea:
@GrahamCampbell Like 492ae3c?

src/NewCommand.php
regiszanandrea: I think it's ok now, right?

src/NewCommand.php
```diff
@@ -81,6 +107,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }, $commands);
         }
 
+        if ($input->getOption('with')) {
```
nunomaduro: Here you can do `if ($with = $input->getOption('with')) {` and replace all `$input->getOption('with')` bellow with `$with`.

src/NewCommand.php
```diff
@@ -243,4 +283,73 @@ protected function findComposer()
...
+     * @return  string
+     */
+    private function getLaravelInstallCommands($packages)
+    {
+        $laravelPackages = explode(',', $packages);
+
+        $packagesInstallCommands = '';
+
+        foreach ($laravelPackages as $laravelPackage) {
+            if (array_key_exists($laravelPackage, self::LARAVEL_PACKAGES_INSTALL_COMMAND)) {
```
nunomaduro: `array_key_exists` is deprecated: https://www.php.net/manual/es/migration74.deprecated.php#migration74.deprecated.core.array-key-exists-objects.

src/NewCommand.php
```diff
@@ -243,4 +283,73 @@ protected function findComposer()
 
         return 'composer';
     }
+
+    /**
+     * @param  string  $packages
+     * @param  string  $composer
+     * @return  string
+     */
+    private function getLaravelPackages($packages, $composer)
```
nunomaduro: This should be `protected` instead.

src/NewCommand.php
```diff
@@ -243,4 +283,73 @@ protected function findComposer()
...
+
+        return $composerPackages;
+    }
+
+    /**
+     * @param  string  $packages
+     * @param  string  $composer
+     * @return  string
+     */
+    private function getLaravelDevPackages($packages, $composer)
```
nunomaduro: This should be `protected` instead.

src/NewCommand.php
```diff
@@ -243,4 +283,73 @@ protected function findComposer()
...
+        }
+
+        return $composerPackages;
+    }
+
+    /**
+     * @param  string  $packages
+     * @return  string
+     */
+    private function getLaravelInstallCommands($packages)
```
nunomaduro: This should be `protected` instead.

src/NewCommand.php
nunomaduro: The only difference between `getLaravelPackages` and `getLaravelDevPackages` is the flag `--dev` and the `list of packages`.

Maybe consider making a generic function like `getPackages($composer , $with, $dev = true)`.

regiszanandrea:
@GrahamCampbell Fixed, the checks are returning ok.

src/NewCommand.php
regiszanandrea: Do you recommend use `isset`?

nunomaduro: Yes.

regiszanandrea:
@nunomaduro I changed the code by your comments. Thanks so much <3

nunomaduro:
@regiszanandrea No worries. I would like to review little more your pull request, but before we do it. Let's make sure @taylorotwell would like to have this on the installer.

taylorotwell:

mattstauffer:
@regiszanandrea If you're still interested in this functionality, it would make sense for us to add it to Lambo. However, I'd recommend we wait until I finish re-writing it in PHP, which I hope to complete in the next few weeks.

regiszanandrea:
@mattstauffer Let's do this 💯

PR #111, opened 2020-02-21
Switch to Github Actions

driesvints:


peter279k:
The GitHub action is failed when executing tests.

It seems that the file permission has been problematic and the installer cannot create project files.

PR #114, opened 2020-04-07
Add a docs command to the CLI tool

djam90:
I'v created this PR to allow any user who has the laravel installer globally installed to be able to run a command such as `laravel docs 5.2` to quickly get to the docs for a given Laravel version.

Perhaps something similar could be introduced in the Laravel core, to be able to run something like `php artisan docs` which would open the docs for a given project. 

I feel like this will be a very useful addition to the laravel installer because most developers will have this package globally installed with composer, and I think many developers still maintain legacy sites for various reasons. It becomes a bit tedious navigating to the docs website and then finding the specific version etc.

I have tried to add valid tests but happy to take feedback on improving them.

Thanks! 😄

### DIFF
--- a/src/DocsCommand.php
+++ b/src/DocsCommand.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace Laravel\Installer\Console;
+
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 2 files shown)


Neilbircumshaw:
Awesome!

PR #123, opened 2020-08-07
Edit the RuntimeException message

parnissolle:
Edit the message sent when extracting the archive to better match the user's request.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -59,8 +59,11 @@ class NewCommand extends Command
 
         $output->writeln('<info>Crafting application...</info>');
 
-        $this->download($zipFile = $this->makeFilename(), $this->getVersion($input))
-             ->extract($zipFile, $directory)
+        $version = $this->getVersion($input);
+        $filename = $this->getFilename($version);
+
+        $this->download($zipFile = $this->makeFilename(), $filename)
+             ->extract($zipFile, $directory, $filename)
              ->prepareWritableDirectories($directory, $output)
              ->cleanUp($zipFile);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Seems like a lot of other changes here.

src/NewCommand.php
```diff
@@ -236,6 +228,29 @@ protected function getVersion(InputInterface $input)
         return 'master';
     }
 
+    /**
+     * Get the filename that should be downloaded.
+     *
+     * @param  string  $version
+     * @return string
+     */
+    protected function getFilename($version)
```
parnissolle, line 237: A method to get the archive filename

src/NewCommand.php
```diff
@@ -133,23 +136,11 @@ protected function makeFilename()
      * Download the temporary Zip to the given file.
      *
      * @param  string  $zipFile
-     * @param  string  $version
+     * @param  string  $filename
      * @return $this
      */
```
parnissolle, line 141: The version is now passed to the `getFilename` method to properly get the required filename

src/NewCommand.php
```diff
@@ -59,8 +59,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $output->writeln('<info>Crafting application...</info>');
```
parnissolle, line 61: We now get the file name from version here, and pass it to `download` and `extract` methods.

src/NewCommand.php
```diff
@@ -162,16 +153,17 @@ protected function download($zipFile, $version = 'master')
...
-    protected function extract($zipFile, $directory)
+    protected function extract($zipFile, $directory, $filename)
     {
         $archive = new ZipArchive;
 
         $response = $archive->open($zipFile, ZipArchive::CHECKCONS);
 
         if ($response === ZipArchive::ER_NOZIP) {
-            throw new RuntimeException('The zip file could not download. Verify that you are able to access: http://cabinet.laravel.com/latest.zip');
+            throw new RuntimeException("The zip file could not download. Verify that you are able to access: http://cabinet.laravel.com/{$filename}");
```
parnissolle, line 166: Then here we are... the purpose to the pull request was to unify the way we render the file name to the user.

PR #126, opened 2020-09-03
Using the assertSame and fix assertSame parameters

peter279k:
# Changed log

- Using the `assertSame` to replace `assertEquals` and it can make assertion equals strict.
- According to the official [PHPUnit doc](https://phpunit.readthedocs.io/en/9.0/assertions.html#assertsame), the `assertSame` parameters are `assertSame(mixed $expected, mixed $actual[, string $message = ''])`.
And parameters should be switched.


driesvints:
Thanks

Issue #128, opened 2020-09-03
Remove command blocks Laravel Installer on Windows PowerShell


volkanmetin:
- Installer Version: 4.0.0

### Description:

When I tried to create a new Laravel project via installer, "rm" method was not recognized and process stopped. Remove command is not working on Windows PowerShell.

Remove command: https://github.com/laravel/installer/pull/124/files#diff-2f97ce541989fd5777bb883c213d1805R61

### Steps To Reproduce:

Run `laravel new blog`

![image](https://user-images.githubusercontent.com/1078613/92173392-ab754e80-ee44-11ea-86b6-feec2bead285.png)

tyler36:
Same issue here.

Lavavel Install: 4.0.0
OS: Win10
Shell: Tried from Powershell, Powershell Core 7.0.3, CMD

ExpDev07:
Getting the same error.

Lavavel Install: 4.0.0
OS: Win10
Shell: Tried both `cmd.exe` and Powershell.

ExpDev07:
I believe this is the line that is causing the issue: https://github.com/laravel/installer/blob/master/src/NewCommand.php#L94.

tyler36:
Confirmed.
Laravel Installer ```4.0.1``` resolves the problem for me.

Thank you.

PR #131, opened 2020-09-04
Fixes https://github.com/laravel/installer/issues/128

ExpDev07:
This pull request fixes https://github.com/laravel/installer/issues/128 by checking the user's operating system and executing the appropriate command for removing directories on that OS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 /vendor
 composer.lock
 phpunit.xml
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea/
 /vendor
 composer.lock
 phpunit.xml
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -91,7 +91,11 @@ class NewCommand extends Command
         ];
 
         if ($directory != '.') {
-            array_unshift($commands, "rm -rf $directory");
+            if (PHP_OS_FAMILY == 'Windows') {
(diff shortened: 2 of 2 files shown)


ExpDev07:
I have to be honest though, I don't 100% get the reason behind this line of code unless the user executes the command with the `--force` option. Otherwise, the code stops at `$this->verifyApplicationDoesntExist($directory)` (https://github.com/laravel/installer/blob/master/src/NewCommand.php#L83) if the project directory already exists, right?

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
peter279k, line 1: Could you explain why this folder is defined on `.gitignore` file and let this be not under Git version control?

Which tool generate this folder?

ExpDev07: It's an IDE folder generated by JetBrains PHP Storm. Let me remove it from .gitignore :)

Bradez:
> I have to be honest though, I don't 100% get the reason behind this line of code unless the user executes the `new` command with the `--force` option. Otherwise, the code stops at `$this->verifyApplicationDoesntExist($directory)` (https://github.com/laravel/installer/blob/master/src/NewCommand.php#L83) if the project directory already exists, right? And if it doesn't exist, then there's no need for that `rm -rf` command to be ran anyways.

I completely agree, looks like it may have been a slight oversight.

driesvints:
Closing this one in favor of https://github.com/laravel/installer/pull/133 which has tests. Thanks.

PR #132, opened 2020-09-04
[4.x] The Laravel Installer requires PHP-7.3 at least

peter279k:
# Changed log

- According to this [line](https://github.com/laravel/installer/blob/master/src/NewCommand.php#L72) of `src/NewCommand` class and the [line](https://github.com/laravel/installer/blob/master/.github/workflows/tests.yml#L16) of `.github/workflows/tests.yml` file,  it should let this installer require `php-7.3` version at least on `composer.json` file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.9",
+        "php": "^7.3",
         "symfony/console": "^4.0|^5.0",
         "symfony/process": "^4.2|^5.0"
     },
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -69,10 +69,6 @@ class NewCommand extends Command
 
         sleep(1);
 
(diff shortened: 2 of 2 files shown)


driesvints:
Think we can remove the warning then as well.

peter279k:
> 
> 
> Think we can remove the warning then as well.

You're  right. these following codes can be removed.

```PHP
        if (version_compare(PHP_VERSION, '7.3.0', '<')) {
            throw new RuntimeException('The Laravel installer requires PHP 7.3.0 or greater. Please use "composer create-project laravel/laravel" instead.');
        }
```

driesvints:
@peter279k can you please add that to the PR?

peter279k:
> 
> 
> @peter279k can you please add that to the PR?

Sure. It's removed :).

driesvints (approved):

Issue #130, opened 2020-09-04
'rm' command not reconized on windows when running with --jet flag and selecting inertia with no teams


ExpDev07:
- Installer Version: 4.0.0

### Description:

The following error is produced when running the command `laravel new test-jet-projec --jet` and selecting Inertia with no teams in a Windows terminal.

### Steps To Reproduce:

1. Open windows terminal.
2. Type `laravel new test-jet-projec --jet` and hit enter.
3. Select option `1` (inertia).
4. Select `n` (no).

If you followed these steps, you should recieve the following error: 
```
'rm' is not recognized as an internal or external command,
operable program or batch file.
```

ExpDev07:
Same issue as https://github.com/laravel/installer/issues/128 it appears like.

PR #133, opened 2020-09-04
[4.x] Fix multiple issues when running on Windows.

### LINKED ISSUE #128: Remove command blocks Laravel Installer on Windows PowerShell

Bradez:
This is an extension of #131 which fixes #128.

This PR also prevents running chmod on Windows (which would equally fail) and only executes the removal of a directory if the `--force`  option has been passed. It also adds tests for for Windows.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ on:
     - cron: '0 0 * * *'
 
 jobs:
-  tests:
+  linux_tests:
 
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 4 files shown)


peter279k:
I think PR #133 and PR #131 can be one PR because issue #133 has the changes from PR #131.

Bradez:
> I think PR #133 and PR #131 can be merged to be the one PR because issue #133 contains the changes from PR #131.

I've gone ahead and opened up ExpDev07/installer/pull/1 so that we can optionally (depending on how the Laravel team wants to approach this) merge this into #131 instead and close this particular issue which is an extension of those fixes, instead of vice versa.

driesvints (approved):

mdpoulter:
I ran into a strange situation where the Laravel installer deleted several files of mine unexpectedly. This PR appears to fix the issue, but I would suggest that a new release is tagged soon so that this doesn't happen to others.

Essentially, having upgraded to v4.0 on my Windows machine, I ran the following in a directory such as `c:\foo bar\baz\`:
```
laravel new untitled --dev
```

However, I also had a directory `c:\foo\` with several files in it too. When the Laravel installer ran the command `rm -rf $directory`, rm evaluated the directory string `c:\foo bar\baz\untitled` as 2 seperate items to delete, specifically,  `c:\foo` and `bar\baz\untitled`. Needless to say, it therefore ended up deleting my `c:\foo\` directory.

As I said, I think this PR fixes the issue but I just wanted to give you a heads up.

Bradez:
@taylorotwell @driesvints can we get a `v4.0.1` tag for the current `HEAD` of `master`? This will at least unblock new users (or those upgrading) who're installing the latest installer.

If there's a possibility that a directory is being deleted, when a space is present in the path (which makes sense due to the lack of quotation marks around the directory), then it's quite a serious issue to push a new release for.

Will also make sure that they're able to install on windows too (since reports of that issue keeps cropping back up due to the lack thereof a new release tag for `^4.0`). #128 #130 #135 #141.

Thanks!

Issue #138, opened 2020-09-04
 Change app.php when installed Jetstream


dleicam:
When installing with JetStream, the following lines should be add in app.php:

`
App\Providers\FortifyServiceProvider::class, 

App\Providers\JetstreamServiceProvider::class,
`

PR #139, opened 2020-09-04
Fix issues (#138)

dleicam:
Add the FortifyServiceProvider and JetstreamServiceProvider to the file "app.php" and only change the .env file if a name is given to the project


taylorotwell:
Those service providers are installed by jetstream:install

PR #142, opened 2020-09-08
APP_URL should be lowercase.

AminulBD:
Currently, The command only fixes the `DB_DATABASE` but the `APP_URL` also needs to be lowercase.


driesvints:
Just don't put an uppercase url in your env variables?

Issue #144, opened 2020-09-08
Installer crashes with combination of --jet and -n


snapey:
- Installer Version: 4.0.1

### Description:
When using non-interactive mode, the installer crashes

PHP Fatal error:  Uncaught TypeError: Argument 2 passed to Laravel\Installer\Console\NewCommand::installJetstream() must be of the type string, null given, called in /Users/mark/.composer/vendor/laravel/installer/src/NewCommand.php on line 116 and defined in /Users/mark/.composer/vendor/laravel/installer/src/NewCommand.php:135

### Steps To Reproduce:

`laravel new myproject --dev --jet -n`

Installer requires decisions about jetstream (livewire/inertia, teams) but there does not seem to be a way to provide these answers when non-interactive

driesvints:
Thanks. I sent in a PR for that: https://github.com/laravel/installer/pull/146

driesvints:
Fixed.

bsienn:
I still have this issue

windows 10
php 7.3

#148

dimawebmaker:
Had same error. 
Reinstall laravel installer helps me.
`
composer global remove laravel/installer

composer global require laravel/installer
`

[source](https://stackoverflow.com/questions/61051679/how-to-update-laravel-installer-laravel-installer-to-latest-version)

bsienn:
@dimawebmaker yes, your provided solution worked. Many thanks for your time.

PR #146, opened 2020-09-08
Provide stack question default

driesvints:
Fixes the `--jet` option when run in non-interactive mode.

Fixes https://github.com/laravel/installer/issues/144


themsaid:
This PR takes care of running the command in non-interactive mood https://github.com/laravel/installer/pull/143

PR #162, opened 2020-10-04
Fix scaffolding in cwd

### LINKED ISSUE #161: Installer claiming application already exists on an empty directory
- Installer Version: 4.0.5

### Description:

If `laravel new` is executed on an empty directory, I would expect the installer to build a new empty Laravel 8 project in the current directory. Instead, it throws an **Application already exists** `RuntimeException`.

Looking at the code, the installer script already checks if the `$directory` is the current working directory, but it fails because `$directory` is not resolved into an absolute path and thus contains "." at the time of execution.

```php
if ((is_dir($directory) || is_file($directory)) && $directory != getcwd())
```

### Steps To Reproduce:

```bash
mkdir ~/test && cd ~/test
laravel new
```

lupinitylabs:
This pull request fixes #161 .

The path in `$directory` is now expanded to an absolute directory using `realpath()`.
Tested on Linux, could not test on Windows (but I hope the CI pipeline will take care of that).

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -180,7 +180,7 @@ class NewCommand extends Command
      */
     protected function verifyApplicationDoesntExist($directory)
     {
-        if ((is_dir($directory) || is_file($directory)) && $directory != getcwd()) {
+        if ((is_dir($directory) || is_file($directory)) && realpath($directory) != getcwd()) {
             throw new RuntimeException('Application already exists!');
         }
     }
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -66,7 +66,7 @@ class NewCommand extends Command
 
         $name = $input->getArgument('name');
 
-        $directory = $name && $name !== '.' ? getcwd().'/'.$name : '.';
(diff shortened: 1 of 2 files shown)


lupinitylabs:
OK, I debugged the script on Windows to see why the tests were failing and fixed another issue where the `$scaffoldDirectory` was not converted to a Windows format path before passing it in to `rd`.  This resulted in a path like `C:\laravel-installer\tests/../tests-output/my-app` being passed to `rd` and then failing.

All tests run fine for me now on PHP 7.3 and 7.4 on windows, I have no idea why the CI tests fail with statusCode 1 for the new test case.

I'll debug some more :-) Any ideas anyone?

lupinitylabs:
As it turns out, the chained cd did not work correctly in all situations.
I also did some refactoring along the way and replaced all hardcoded forward slashes in pathes with DIRECTORY_SEPERATOR. This also saves us the ugly `preg_replace`.

It works correctly for me locally, the CI tests still fail, however. Anyone got an idea why?

driesvints:
Heya. I'm sorry but this isn't supported and this pr also seems to break tests in Windows.

PR #168, opened 2020-10-29
[4.x] PHP 8 Support

driesvints:


.github/workflows/tests.yml
```diff
@@ -7,41 +7,42 @@ on:
...
 
   windows_tests:
 
     runs-on: windows-latest
     strategy:
       fail-fast: true
       matrix:
-        php: [7.3, 7.4]
+#        php: [7.3, 7.4, ^8.0]
+        php: [^8.0]
```
GrahamCampbell, line 45: carrot is a typo i assume?

GrahamCampbell (changes requested):

.github/workflows/tests.yml
driesvints: yep thanks

nuernbergerA:
Hey @driesvints 
#170 should fix it.

driesvints:
@taylorotwell Although that this library doesn't explicitly requires it, I've added `ext-fileinfo` to its dependencies which will allow us to catch the snag from above in the future more easily.

crynobone:
I don't think there is a need for that. From what I see the ubuntu inage has php installed with fileinfo by default while windows doesn't seem to have it. 

I seen this issue all all tests with Laravel 9 at the moment (when running on windows CI)

driesvints:
Okay I've removed it again. We'll see how it goes.

PR #175, opened 2020-11-09
Adds --breeze option

nunomaduro:
This pull request adds the `--breeze` option that behaves just like the `--jet` option, but for Laravel Breeze.

```bash
laravel new foo --breeze
```


taylorotwell:
I'm actually planning to sort of steer people away from these flags for undisclosed reasons 😄  ...

PR #176, opened 2020-11-11
Optimize file import of Composer

guanguans:
> Optimize file import of Composer.


taylorotwell:
Current code seems fine to me.

PR #177, opened 2020-11-17
Fix Application already exist error when creating a new app on curren…

### LINKED ISSUE #161: Installer claiming application already exists on an empty directory

TK0423:
This pull request fixes #161 .

A similar PR was rejected claiming that current directory is not supported, however the name argument is still label as optional  in [NewCommand.php#L27](https://github.com/laravel/installer/blob/master/src/NewCommand.php#L27) allowing for **laravel new**  to be run on the current directory

In previous versions **laravel new**  and **laravel new .** were valid options, and  [NewCommand.php#L73](https://github.com/laravel/installer/blob/master/src/NewCommand.php#L73) still considers that name may be empty  or . 
it's just incorrectly setting the directory to . instead of the current path

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -70,7 +70,7 @@ class NewCommand extends Command
 
         $name = $input->getArgument('name');
 
-        $directory = $name && $name !== '.' ? getcwd().'/'.$name : '.';
+        $directory = $name && $name !== '.' ? getcwd().'/'.$name : getcwd();
 
         $version = $this->getVersion($input);
 
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -43,7 +43,7 @@ class NewCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $installJetstream = $input->getOption('jet') ||
-            ($input->getOption('prompt-jetstream') && (new SymfonyStyle($input, $output))->confirm('Would you like to install the Laravel Jetstream application scaffolding?', false));
+                            ($input->getOption('prompt-jetstream') && (new SymfonyStyle($input, $output))->confirm('Would you like to install the Laravel Jetstream application scaffolding?', false));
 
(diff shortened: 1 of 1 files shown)


driesvints:
This will break checks further down where `.` is being checked upon.

src/NewCommand.php
```diff
@@ -43,7 +43,7 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $installJetstream = $input->getOption('jet') ||
-                            ($input->getOption('prompt-jetstream') && (new SymfonyStyle($input, $output))->confirm('Would you like to install the Laravel Jetstream application scaffolding?', false));
+            ($input->getOption('prompt-jetstream') && (new SymfonyStyle($input, $output))->confirm('Would you like to install the Laravel Jetstream application scaffolding?', false));
```
driesvints, line 46: Please revert all these styling changes in this PR.

TK0423:
> ed u

you are right, I forgot about the --force parameter, I added a commit to account for this additional checks
Instead of checking $directory to see if it can be forced, we use $name, $name must be not empty and != '.'

src/NewCommand.php
TK0423: done

taylorotwell:
We should just make name required.

PR #191, opened 2021-03-12
Improve Git push

driesvints:
After talking to some of my GitHub contacts they gave me a tip for a safer way to do a `git push` after creating the GitHub repo. Doing so will allow us to remove the note about the `gh_protocol` in the docs as it won't matter anymore what protocol you use.

PR #192, opened 2021-03-16
Revert "Improve Git push"

driesvints:
Reverts laravel/installer#191


driesvints:
Reverting because this will break for non-WSL Windows installs.

PR #195, opened 2021-03-18
Change the organisation flag to be optional

djekl:
@driesvints Thanks for including this, but shouldn't this be optional?


driesvints:
No the value is required when you use this flag.

djekl:
Ahh, completely misunderstood. Sorry!

PR #199, opened 2021-04-24
Add support for oldest versions of Git

HajMo:
This PR will add support for git version less than 2.28.0. Which is the only version that has the `-b` option for creating a repo during the initialization.
This will help Linux users which is by default to have Git version less than 2.28.0.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -211,9 +211,10 @@ class NewCommand extends Command
         $branch = $input->getOption('branch') ?: 'main';
 
         $commands = [
-            "git init -q -b {$branch} .",
+            'git init -q',
             'git add .',
             'git commit -q -m "Set up a fresh Laravel app"',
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Using git older than 2.30.2 is not recommended because of known security vulnerabilities.

HajMo:
Agree. But for people who just use what is installed by default now they can't use `--git` option in laravel installer. Even github instructions for new repositories doesn't contain the new `-b` option

GrahamCampbell:
Isn't that a good thing? WIll make them upgrade and not have a vulnerable version of git.

driesvints (approved):

driesvints:
@GrahamCampbell It's okay. I don't really mind this. More people will be able to use the `git` flag this way. We shouldn't require what people use as a minimum git just for this feature.

PR #200, opened 2021-05-14
Change default `APP_NAME` to the project name

HajMo:
This pull request will make the `APP_NAME` variable hold the project name instead of the default value (Laravel).


driesvints:
There's some commits in here which don't belong here. Please re-create the PR with a fresh commit history. Thanks

PR #207, opened 2021-08-11
Respect git global config for a default branch

SamAsEnd:
Git provides a global configuration to override the default branch when a new repository is initialized.
Even though laravel/installer provides a way to override the default branch name, laravel/installer does respect git global configurations.

[init.defaultBranch:- Allows overriding the default branch name e.g. when initializing a new repository.](https://git-scm.com/docs/git-config#Documentation/git-config.txt-initdefaultBranch)

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -27,7 +27,7 @@ class NewCommand extends Command
             ->addArgument('name', InputArgument::REQUIRED)
             ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
             ->addOption('git', null, InputOption::VALUE_NONE, 'Initialize a Git repository')
-            ->addOption('branch', null, InputOption::VALUE_REQUIRED, 'The branch that should be created for a new repository', $this->defaultBranch('main'))
+            ->addOption('branch', null, InputOption::VALUE_REQUIRED, 'The branch that should be created for a new repository', $this->defaultBranch())
             ->addOption('github', null, InputOption::VALUE_OPTIONAL, 'Create a new repository on GitHub', false)
             ->addOption('organization', null, InputOption::VALUE_REQUIRED, 'The GitHub organization to create the new repository for')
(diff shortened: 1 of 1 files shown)


SamAsEnd:
## git config --global
Git

## laravel/installer respecting git config --global
laravel/installer

SamAsEnd:
@taylorotwell This is not a "feature" I can release as a package.

Laravel installer SHOULD respect my git global configuration for my "master/main" branch.

Otherwise I'm forced to always pass the branch argument which is not nessasery.

Please, respect git configuration
@driesvints @themsaid

taylorotwell:
How does this PR behave if Git isn't installed?

src/NewCommand.php
```diff
@@ -143,6 +143,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+     * @return string
+     */
+    protected function defaultBranch($default)
+    {
+        $process = new Process(['git', 'config', '--global', 'init.defaultBranch']);
+        $process->run();
+
+        $output = trim($process->getOutput());
+
+        return $process->isSuccessful() && $output ? $output : $default;
```
driesvints, line 158: I'd consider removing the parameter from the signature and simply setting the `main` default here.

SamAsEnd:
> How does this PR behave if Git isn't installed?

It will default to `main` since the process will fail.

src/NewCommand.php
SamAsEnd: @driesvints Check it out

src/NewCommand.php
```diff
@@ -143,6 +143,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+     * @return string
+     */
+    protected function defaultBranch()
+    {
+        $process = new Process(['git', 'config', '--global', 'init.defaultBranch']);
+        $process->run();
+
+        $output = trim($process->getOutput());
+
+        return $process->isSuccessful() && $output ? $output : 'main';
```
SamAsEnd, line 158: @taylorotwell If the process is not successful, it will default to `main`

PR #208, opened 2021-09-25
Add valet option to link and secure new installation directory

tomhrtly:
This PR adds a new "valet" option to the installer which can be used to automatically link and secure the new installation for devs using Valet.

It does not break any existing features and is optional to use. It saves a bit of time as you do not need to type out the Valet commands each time you create a new installation.


taylorotwell:

PR #219, opened 2022-01-11
Maintain current functionality with gh repo create rewrite

ArnaudLier:
Flag --confirm / -y has been deprecated, passing any argument skips confirmation prompt.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -288,7 +288,7 @@ class NewCommand extends Command
         $branch = $input->getOption('branch') ?: $this->defaultBranch();
 
         $commands = [
-            "gh repo create {$name} -y {$flags}",
+            "gh repo create {$name} {$flags}",
             "git -c credential.helper= -c credential.helper='!gh auth git-credential' push -q -u origin {$branch}",
         ];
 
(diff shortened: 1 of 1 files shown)


driesvints:
Where is the news about this? I can't find anything in the GitHub CLI release notes.

ArnaudLier:
I didn't check the release notes but here is the pull requests that deprecates the flags: https://github.com/cli/cli/pull/4578
Maybe we should add the --source flag to maintain current functionality (the `origin` remote isn't added by default anymore).

driesvints (approved):

driesvints:
Thanks @ArnaudLier

src/NewCommand.php
```diff
@@ -288,7 +288,7 @@ protected function pushToGitHub(string $name, string $directory, InputInterface
         $branch = $input->getOption('branch') ?: $this->defaultBranch();
 
         $commands = [
-            "gh repo create {$name} -y {$flags}",
+            "gh repo create {$name} --source=. {$flags}",
```
driesvints, line 291: Why did you added this?

ArnaudLier: To maintain current functionality, the installer can't push the git log if there is no origin remote:
> For example, running gh repo create myrepo --public in a local repository will no longer add a remote. It assumes you just want to create a new repository on GitHub. To create a remote repo from your local git repo, use gh repo create myrepo --public --source=. instead

PR #225, opened 2022-04-27
Add option to support SSR if jetstream and inertia selected

danyelkeddah:
Add ability to support SSR if Jetstream scaffolded with inertia stack


taylorotwell:
I would prefer people not use the `--jet` flag. I don't document it anymore. Please just install Jetstream separetly.

PR #231, opened 2022-06-18
Add node-packages option to determine if installer should install node packages or not

Kamona-WD:
If you provided `--node-packages` option the installer will  install `node_modules`
for you after `laravel` installation .

```sh
laravel new my-project --node-packages
```
new-laravel-project

and with `jetstream`

```sh
laravel new my-project --jet
```

new-laravel-project-with-jet


taylorotwell:
I think I'm fine going without adding more options to this command.

PR #232, opened 2022-06-28
Use build command instead of dev command

driesvints:
This fixes support for the new Vite pipeline.


timacdonald:
Nice catch @driesvints !

jessarcher:
Thanks, @driesvints!

Is it worth putting those in the Jetstream installer instead so that it's the same experience whether you use the Laravel installer or install Jetstream manually?

I'm also wondering whether Breeze should be in here as well? That might be part of a larger conversation around the various ways of installing Laravel and starter kits.

PR #236, opened 2022-08-01
Adds pest support for jetstream

mozex:
Before this PR you could only install jetstream with the PHPUnit test suite, this PR adds the option to choose the test suite in the installing process.

Preview

It is possible to add a test suite as an option too, for example:
`laravel new example-app --jet --test-suite=pest`

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -33,6 +33,7 @@ class NewCommand extends Command
             ->addOption('jet', null, InputOption::VALUE_NONE, 'Installs the Laravel Jetstream scaffolding')
             ->addOption('stack', null, InputOption::VALUE_OPTIONAL, 'The Jetstream stack that should be installed')
             ->addOption('teams', null, InputOption::VALUE_NONE, 'Indicates whether Jetstream should be scaffolded with team support')
+            ->addOption('test-suite', null, InputOption::VALUE_OPTIONAL, 'The Jetstream test suite that should be installed')
             ->addOption('prompt-jetstream', null, InputOption::VALUE_NONE, 'Issues a prompt to determine if Jetstream should be installed')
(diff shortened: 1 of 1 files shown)


driesvints:
Best to convert it to `--pest` which is more readable.

src/NewCommand.php
```diff
@@ -170,13 +173,14 @@ protected function defaultBranch()
      * @param  \Symfony\Component\Console\Output\OutputInterface  $output
      * @return void
      */
-    protected function installJetstream(string $directory, string $stack, bool $teams, InputInterface $input, OutputInterface $output)
+    protected function installJetstream(string $directory, string $stack, bool $teams, string $testSuite, InputInterface $input, OutputInterface $output)
```
driesvints, line 176: No need to pass this through here. Check the `--pest` option inside this method.

driesvints (changes requested):

src/NewCommand.php
mozex: Ok, I will push the new changes.

mozex: Wouldn't it be better to ask whether users want to choose Pest or not? if I move the option check inside the installJetstream method, we shouldn't ask a question anymore.
the reason that I checked the option in there was to ask all the questions on the first of install.

driesvints: They can choose Pest by providing `--pest`. This makes the command equal to how Jetstream install works. There's no need for a question.

mozex: Done.

driesvints (approved):

taylorotwell:
I'm trying to avoid adding more options here. We don't actually document installing Jetstream in this way anymore.

PR #237, opened 2022-08-01
No teams patch

SketchNI:
This PR creates a new option `--no-teams|-s` to explicitly prevent being asked to create a team during project initialisation.


driesvints:
Thanks. Right now, we are not accepting new flags for this command.

PR #238, opened 2022-08-02
Simplify push to GitHub

driesvints:
GitHub has added a new `--push` flag which simplifies the process of pushing the HEAD branch to the newly created repo.

See https://github.com/cli/cli/issues/3209#issuecomment-1161572256

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -289,8 +289,7 @@ class NewCommand extends Command
         $branch = $input->getOption('branch') ?: $this->defaultBranch();
 
         $commands = [
-            "gh repo create {$name} --source=. {$flags}",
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -289,8 +289,7 @@ protected function pushToGitHub(string $name, string $directory, InputInterface
         $branch = $input->getOption('branch') ?: $this->defaultBranch();
```
fgaroby, line 289: Isn't this variable `$branch` (or even the whole line) useless now?

driesvints: I removed this line: https://github.com/laravel/installer/commit/ea0b3b6e5f9d6486bdf05342581224b007b817c0

PR #241, opened 2022-08-09
Replace APP_URL in .env.example

RhysLees:
This PR aims to replicate the same behaviour as the .env replacement for APP_URL in the .env.example file.


taylorotwell:
No plans to change this atm.

PR #244, opened 2022-09-04
Replace PHP version in composer.json

DvDty:
The combination between having the default PHP version in `composer.json`, which is `laravel/laravel`'s lowest supported and PHPStorm, creates an illusion that you can't use newer features, unless you change it yourself (which I haven't seen mentioned in the official documentation or the starting guides around the web).
Here is an example - on a new Laravel installation (via `sail` with PHP v8.1.9) I'm getting this "error":

Screenshot from 2022-09-03 16-40-26

PHPStorm gives me the option to switch the version, but for newer folks, which would be the people most likely to have this issue, it may not be clear what does this mean. On top of that, PHPStorm gives me this option even if my PHP version is below 8.1, meaning I would later get runtime errors if I chose to switch and use the newer features.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -431,7 +431,7 @@ class NewCommand extends Command
     /**
      * Replace the default PHP version in composer.json.
      *
-     * @param string $directory
+     * @param  string $directory
      * @return void
      */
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #247, opened 2022-10-28
Added arguement for installing Auth stack Breeze directly

cart53:
Laravel is not used to generate static site as we all know. So if we give the option to install Breeze directly while installing laravel with "laravel new *" command, it will be easy for all developers who later write the command


taylorotwell:
I'm not really adding new options to this package.

PR #249, opened 2022-10-29
Add --pest option to use Pest PHP when installing Jeststream

sikessem:
Available options do not include --pest which installs the PestPHP test suite instead of the default PHPUnit test suite.
However, this option is specified in [the Jetstream installation documentation](https://jetstream.laravel.com/2.x/installation.html#installing-jetstream).

Jetstream installation documentation


taylorotwell:
I don't add new options to the Laravel installer. I recommend installing Jetstream after creating the Laravel projects and having the full power of Jetstream options.

PR #251, opened 2023-02-03
[4.x] Adds `--pest` to use Pest

nunomaduro:
This pull request does two things:

1. First, it adds the `--pest` option, that may be used to create a new project with Laravel using Pest PHP.
2. Second, it adds the `--pest` option (or question when not provided, similar to `--stack`), that may be used to create a new Jetstream using Pest PHP.


driesvints:
Can we also let the install command fail when composer cannot resolve dependencies? Like in the PR you fixed here? https://github.com/laravel/jetstream/pull/1240

PR #252, opened 2023-02-07
feature: allow default config to be set, and used

benjam-es:
Signed-off-by: Ben James <in@benjam.es>

The idea was from a tweet by Freek https://twitter.com/freekmurze/status/1623002154259587072

Where users might prefer pest as their `laravel new` default, or may want jet stream with teams on every new app

This can be done with a new configure command, or by using a `save-defaults` option on the `laravel new` command.

There is a list of saveable options. The options are saved in a config json file

Let me know if the proof of concept is worth continuing, and any suggestions for changes.


taylorotwell:

PR #256, opened 2023-02-15
Add an option to install PHPUnit

likeadeckofcards:
If you try to run the `laravel new --jet` command without the capability to use the interactive form, you will get an error that the testing framework is not provided. Also, adding this a mirror to the `--pest` flag. Since this is the default behavior, adding the `--pest` flag will result in the pest framework being selected.


jrd-lewis:
@driesvints Didn't you just say you're not adding new flags to the installer? This is a new flag.

PR #266, opened 2023-07-01
Add `--ci` option for GitHub Actions integration

osbre:
This PR adds a `--ci` option which includes a fully-functioning GitHub Actions configuration to test the Laravel application with dependency caching, assets compilation and in-memory SQLite database.

I think it would be nice to have this out-of-box, at least for GitHub Actions.

*improving developer experience ✨🚀*

### Notes
Naming this option  `--ci` instead of `--github-actions` to make it extendable with other ci systems or configurations.

Some future ideas:

- `--ci=github-sqlite`
- `--ci=github-mysql`
- `--ci=gitlab-pgsql`

Thoughts?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
     "require": {
         "php": "^7.3|^8.0",
         "symfony/console": "^4.0|^5.0|^6.0",
-        "symfony/process": "^4.2|^5.0|^6.0"
+        "symfony/process": "^4.2|^5.0|^6.0",
+        "symfony/filesystem": "^4.0|^5.0|^6.0"
     },
     "require-dev": {
         "phpstan/phpstan": "^1.10",
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -10,6 +10,7 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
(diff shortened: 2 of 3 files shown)


driesvints:
I think this idea is pretty neat. Let's see what @taylorotwell says.

driesvints:
I was first thinking we could ship with a default workflow file in the Laravel skeleton repository itself but on second thought not everyone might use github actions. One concern is this file getting out of sync with the skeleton so that's an argument again to include it in the skeleton by default.

osbre:
@driesvints I don't think this file needs to be synced with the one in the skeleton, because they are different.
https://github.com/laravel/laravel/blob/10.x/.github/workflows/tests.yml
(the one in the skeleton has scheduling and multiple PHP version enabled, don't think people need that in their projects)

And If you look at the file in this PR, it is a more complete version with caching, assets compilation and env variables.
https://github.com/osbre/installer/blob/feat/ci-option/stubs/ci/github-actions.yml

driesvints:
> I don't think this file needs to be synced with the one in the skeleton, because they are different.

I'm not talking about that one. I'm talking about the script from this PR to be moved into the skeleton one. The one you refer to is just to verify the skeleton works on all supported PHP versions.

osbre:
@driesvints I wouldn't mind having it directly in the skeleton, but I guess some people would be against it. These are people who use GitLab, BitBucket or Gitea.

That's why I think it fits perfectly into the `installer` because the arguments in the installer are optional and are a choice.

taylorotwell:
I find this a bit opinionated with the asset upload, compilation, etc.

PR #270, opened 2023-08-08
Show the directory name at the end of the installer

jessarcher:
This PR updates the final output of the installer to include the chosen directory name:

image

This might not be an issue for most, but I often forget the directory name I chose, especially if I switched away while the installer runs.

The Vue installer does something similar (including the name twice), but I opted not to provide the `cd` instruction because it looked a little weird on its own (we don't need to tell them to `composer install`) and developers may want to open the directory in their editor instead of "cd-ing".

image

I also wondered whether it could be worth adding some "Next Steps" including links to the docs on running development servers and using starter kits. Thoughts?

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -193,9 +193,7 @@ class NewCommand extends Command
                 $output->writeln('');
             }
 
-            $output->writeln("  <bg=blue;fg=white> INFO </> Application ready in <options=bold>[{$name}]</>.".PHP_EOL);
-
-            $output->writeln('  <bg=blue;fg=white> INFO </> <options=bold>Build something amazing.</>'.PHP_EOL);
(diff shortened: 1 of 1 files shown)


driesvints (approved):

src/NewCommand.php
```diff
@@ -193,7 +193,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 $output->writeln('');
             }
 
-            $output->writeln('  <bg=blue;fg=white> INFO </> Application ready! <options=bold>Build something amazing.</>'.PHP_EOL);
+            $output->writeln("  <bg=blue;fg=white> INFO </> Application ready in <options=bold>[{$name}]</>.".PHP_EOL);
```
nunomaduro, line 196: Still feel that the `Build something amazing!` could be the same line. Something like this:

```
[INFO] Application ready in [./my-application]. Build something amazing!
```

nunomaduro (changes requested):

PR #273, opened 2023-08-16
Updated to use the Process constructor to install Pest 

rajabilal555:
For Issue #272
Pest failed to install in Windows PowerShell when selecting no Starter Kit on the initial prompt.

+ Added two new functions from the laravel\breeze repo
  - requireComposerPackages
  - removeComposerPackages
+ Updated the `installPest` function to use `requireComposerPackages` instead
+ Now we can use the `[require/remove]ComposerPackages` functions which handle escaping commands through the Process constructor.

PR #274, opened 2023-08-17
Updated to use the Process constructor to install Pest and use `Composer` helper class

crynobone:
* Continue and replace #273 PR
* Use new helper methods from laravel/framework#48096
* Use `Process` `$cwd` variable to run command from path instead of using `chdir()`


driesvints:
Marking this as draft until Laravel v10.20 is out

PR #275, opened 2023-08-18
Add ability to select default database connection

crynobone:
* Allow selecting `sqlite`, `mysql` or `pgsql` driver using Laravel Prompts
* Default to `mysql` if accessed via `--no-interaction`.
* Update `DB_CONNECTION` to the correct database driver. (SQLite database doesn't update `.env.example`)
* Comment other `DB_*` variables when using `sqlite`.
* Update `DB_PORT` value based on selected database driver.


crynobone:
> Wouldn't you typically change the port on Postgres?

Added script to change default port, and allow selecting MSSQL

PR #284, opened 2023-09-22
Use prompts non-interactive mode instead of Symfony fallbacks

jessarcher:
This PR updates the Laravel installer to use Prompts' new "non-interactive mode" instead of relying on the Symfony fallback.

This does not change the behaviour when using the `-n` argument, as the command will never call the `interact` hook, which is the only place we currently use Prompts in the installer. It only changes the output when there are missing required arguments and the input is not a TTY (such as in CI).

|Before|After|
|-|-|
|image|image|

This change is pretty insignificant for the installer, but it paves the way for https://github.com/laravel/prompts/pull/57, allowing us to remove the fallback code from this repo and any other packages/projects that can't or don't rely on `laravel/framework` to configure a Windows fallback.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^8.1",
         "illuminate/filesystem": "^10.20",
         "illuminate/support": "^10.20",
-        "laravel/prompts": "^0.1",
+        "laravel/prompts": "^0.1.9",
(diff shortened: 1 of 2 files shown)


jessarcher:
This can wait until Prompts `0.1.9` is released.

PR #287, opened 2023-10-04
Remove extra DB_DATABASE

ConnySjoblom:
When creating a new Laravel project using the installer, and selecting sqlite as database the .env file `DB_DATABASE` row ends up being commented out twice:
```
# DB_PORT=3306
# # DB_DATABASE=laravel
# DB_USERNAME=root
```

This PR fixes that by removing the extra line in `$defaults`

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -248,7 +248,6 @@ class NewCommand extends Command
         }
 
         $defaults = [
-            'DB_DATABASE=laravel',
             'DB_HOST=127.0.0.1',
             'DB_PORT=3306',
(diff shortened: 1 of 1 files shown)


driesvints (approved):

crynobone:
Where exactly does it cause `#` to be added twice?

ConnySjoblom:
> Where exactly does it cause `#` to be added twice?

https://github.com/laravel/installer/blob/master/src/NewCommand.php#L262

It loops through all of the defaults, and appends the `#`

crynobone (approved):

PR #291, opened 2023-11-09
add telescope install support

umairparacha00:
This pull request adds support for the Laravel Telescope for both:
- Default installation
- Local only installation

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -120,8 +120,9 @@ class NewCommand extends Command
             match (select(
                 label: 'Would you like to install Laravel Telescope?',
                 options: [
-                    'telescope' => 'Laravel Telescope',
-                    'telescope-local' => 'Laravel Telescope (local)',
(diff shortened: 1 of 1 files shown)


driesvints:
I don't think we should be adding all of the first party packages as flags to the installer like this.

jessarcher:
I kinda like the idea of prompting for some first-party packages, but I reckon a multi-select would be better rather than confirmation prompts. It could even replace the starter-kit selection with a validation error to prevent selecting Jetstream _and_ Breeze. In addition to convenience, it would help bring awareness to available packages. On the other hand, it could make the installer overwhelming.

umairparacha00:
To keep the installer minimal, we can just add the option to install telescope instead of confirmation prompt.

umairparacha00:
Now, I have removed the confirmation prompt, and only when the user passes the option prompt will show up to about the installation type.

PR #292, opened 2023-11-09
Add mariadb installation option

Jubeki:
Laravel 11 will support a new Maria DB Configuration in `config/database.php` (due to slimming the skeleton this will be moved to the framework).

A developer should be able to select the option `mariadb` during installation. Which will fallback to `mysql` in Laravel 10 and will use the `mariadb` configuration in Laravel 11.

~Note: This PR will be draft until the slimmed skeleton is merged. Because the Slimmed skeleton branch is not working, I can not verify if everything works as intended.~

Works on the master branch and the 10.x branch

Tested as follows:
Replace the file in `~/.composer/vendor/laravel/installer/src/NewCommand.php`
Run `laravel new l11 --dev` and select database mariadb => `.env` should contain mariadb
Run `laravel new l10` and select database mariadb => `.env` should contain mysql

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -661,7 +661,7 @@ class NewCommand extends Command
     protected function getVersion(InputInterface $input)
     {
         if ($input->getOption('dev')) {
-            return 'dev-master';
+            return 'dev-slim-skeleton-11.x';
         }
 
         return '';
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -635,7 +661,7 @@ protected function canResolveHostname($hostname)
     protected function getVersion(InputInterface $input)
     {
         if ($input->getOption('dev')) {
-            return 'dev-master';
+            return 'dev-slim-skeleton-11.x';
```
Jubeki, line 664:
```suggestion
            return 'dev-master';
```

This needs to be reverted, once the slim-skeleton branch is merged.

PR #295, opened 2023-12-05
[develop] Adds Laravel 11 support

nunomaduro:
This pull request adds Laravel 11 support to the installer and is awaiting the completion of the starter kits - this will allow for testing with their respective major versions.


driesvints:
@nunomaduro we don't need to do a new major version. This can just go into 5.x

PR #300, opened 2024-01-05
[5.x] Added default value on selection question while creating new application

bigship-prashant:
This pull request aims to enhance the Laravel installer by introducing default values for selection questions during the process of creating a new application. The addition of default values streamlines the installation process, making it more user-friendly and efficient.

With this update, I've implemented default values for certain selection questions that users encounter while creating a new Laravel application. These defaults are thoughtfully chosen to align with best practices and common preferences, ensuring a smoother on-boarding experience for both new and experienced users.

By providing default values, users can simply accept the pre-filled options if they align with their requirements. This significantly reduces the number of manual inputs, making the installation process faster and more straightforward.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -101,6 +101,7 @@ class NewCommand extends Command
                     'breeze' => 'Laravel Breeze',
                     'jetstream' => 'Laravel Jetstream',
                 ],
+                default: 'none',
             )) {
                 'breeze' => $input->setOption('breeze', true),
                 'jetstream' => $input->setOption('jet', true),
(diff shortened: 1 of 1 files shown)


driesvints:
Aren't values already selected by default? Afaik this option is only to select specific ones. I can always just hit enter and it uses the one that's currently selected.

bigship-prashant:
> Aren't values already selected by default? Afaik this option is only to select specific ones. I can always just hit enter and it uses the one that's currently selected.

Nope, I cannot use it on my windows 11 pc. If you use Mac or Linux, I cannot tell but in windows we need to type. So this PR is for reducing that

PR #308, opened 2024-01-24
Check for latest version

### LINKED ISSUE #306: Add check to installer to see if you're on the latest version
And if not, prompt the user to update.

Jubeki:
Closes #306 

This prompts the user to update the Laravel Installer before continuing.

How it looks like when the update will be installed (composer messages are wrong because the Laravel Installer Version was changed manually, and therefor Composer does not update anything)

Screenshot 2024-01-25 at 10 40 03

How it looks like when you continue:

Screenshot 2024-01-25 at 10 40 24

To test this you can to the following:

0. Make sure Laravel Installer is up to date with `composer global update`
1. Copy the content of `src/NewCommand.php` in this PR
2. Paste it into `~/.composer/vendor/laravel/installer/src/NewCommand.php` 
3. Change the version in `~/.composer/vendor/laravel/installer/bin/laravel` into `5.3.0`
4. Run `laravel new test --no-interaction`

To test if the check is not run, you can simply change the version back to `5.4.0`.

### Considerations

There is also the question how you want to handle downtime with Github. Probably then assuming you are on the latest version?

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -843,7 +843,7 @@ class NewCommand extends Command
             $output->write('    '.$line);
         });
 
-        if($this->existsNewerInstallerVersion()) {
+        if ($this->existsNewerInstallerVersion()) {
             $output->writeln('  <bg=green;fg=white> NEW VERSION </> A newer version of Laravel Installer is available. To update simply run <options=bold>`composer global update`</>'.PHP_EOL);
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Going to table this for now.

driesvints:
Sorry @Jubeki, this was my bad. This won't work with situations where the installer is installed using Herd, etc. Thanks for your attempt.

Jubeki:
@driesvints I initially had only printed that a new version is available, would this instead be better:

PR #319, opened 2024-03-11
Remove Laravel 11 Checks

Jubeki:
Tomorrow when Laravel 11 is released, the default branch in https://github.com/laravel/laravel will change.

Therefor the installer will only pull the `11.x` or `master` branch.
The checks are no longer necessary.

I opted for leaving the version check with an version parameter.
(This can also be removed if wanted).

It would probably also best to mention updating the Laravel Installer Version in the docs. (will send a PR for that)


skor235:
I have Laravel 5.6.0 installer. When I run "laravel new project", the installer installs laravel version 11.0.7 by default. is it possible to specify the installation of version 10?

Jubeki:
You can not use the installer to create a Laravel 10 Application, but you can do the following:

To install with breeze:

```bash
composer create-project laravel/laravel:^10.0 example-app
cd example-app
composer require laravel/breeze
php artisan breeze:install
```

To install with jetstream:

```bash
composer create-project laravel/laravel:^10.0 example-app
cd example-app
composer require laravel/jetstream
php artisan jetstream:install
```

Afterwards configure your database accordingly to your environment.

skor235:
I'm aware of that, but the installer is more convenient. I noticed that the latest version of the installer supporting Laravel 10 is version 5.2.1. is it possible to install version 5.2.1 instead of 5.6.0

Jubeki:
The installer always pulls the latest branch from https://github.com/laravel/laravel which in this case is `11.x` with the flag `--dev` you can fetch the master branch.

Event if you are using 5.2.1 you are still getting the 11.x branch.

The version 5.6 is documented, so that you have the latest features available for configuring a new Laravel 11 Application.

(It is generally discouraged to create a Laravel 10 Application, when Laravel 11 is already released)

PR #322, opened 2024-03-12
Add --ver special version option 

sehirger:
Add special version argument

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -44,6 +44,7 @@ class NewCommand extends Command
             ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
             ->addOption('git', null, InputOption::VALUE_NONE, 'Initialize a Git repository')
             ->addOption('branch', null, InputOption::VALUE_REQUIRED, 'The branch that should be created for a new repository', $this->defaultBranch())
+            ->addOption('ver', null, InputOption::VALUE_REQUIRED, 'For special laravel version,e.g --ver=10.x')
             ->addOption('github', null, InputOption::VALUE_OPTIONAL, 'Create a new repository on GitHub', false)
             ->addOption('organization', null, InputOption::VALUE_REQUIRED, 'The GitHub organization to create the new repository for')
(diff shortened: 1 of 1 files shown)


Jubeki:
This is probably not recommended, due to file changes in the skeleton.

You can instead do the following:

```bash
composer create-project laravel/laravel:^10.0 example-app
```

crynobone:
@Jubeki is correct. The installation commands is intended for Laravel 11 only.

sehirger:
> This is probably not recommended, due to file changes in the skeleton.
> 
> You can instead do the following:
> 
> ```shell
> composer create-project laravel/laravel:^10.0 example-app
> ```

If you use composer to create a special version, you will encounter some difficulties in front-end configuration, such as TS support, brezz installation, etc.

Jubeki:
The installer is a convenience for the latest stable Laravel Version (and the next upcoming major version).

You can always install an older Laravel version using composer and then manually install e.g. Breeze with `composer require laravel/breeze` and then `php artisan breeze:install` (though I don't know why you would want to create a Laravel 10 instead of Laravel 11 Application).
This should result in a similar experience.

Same with jestream:
```bash
composer require laravel/jetstream
php artisan jetstream:install
```

Afterwards only some database adjustments in the `.env`, migrate and you are also ready to go.

sehirger:
> The installer is a convenience for the latest stable Laravel Version (and the next upcoming major version).
> 
> You can always install an older Laravel version using composer and then manually install e.g. Breeze with `composer require laravel/breeze` and then `php artisan breeze:install` (though I don't know why you would want to create a Laravel 10 instead of Laravel 11 Application). This should result in a similar experience.
> 
> Same with jestream:
> 
> ```shell
> composer require laravel/jetstream
> php artisan jetstream:install
> ```
> 
> Afterwards only some database adjustments in the `.env`, migrate and you are also ready to go.

I know the way you suggested to create the project, but from the perspective of convenience and speed, I think it is necessary to add the features of optional version. My above modification may be simple and crude, but I hope everyone can find a good way to add the features. Features that specify version to create Laravel projects.

driesvints:
No plans for this as the installer is only intended to be used with the latest version of the framework.

Issue #331, opened 2024-04-11
breeze installs breeze without --dev flag


abumosaab:
Please pardon my ignorance as this is not an issue. I just noticed that `laravel/installer` requires `laravel/breeze` without --dev  flag. 
Not be a big deal. I might be missing somthing. But docs mentions using of --dev.

PR #332, opened 2024-04-11
Install Breeze with dev flag

driesvints:
Breeze is actually recommended to be installed as a dev dependency because it only scaffolds files. Ideally we'd also remove the dependency again after scaffolding Breeze.

https://laravel.com/docs/11.x/starter-kits#laravel-breeze-installation

See https://github.com/laravel/installer/issues/331

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -394,7 +394,7 @@ class NewCommand extends Command
     protected function installBreeze(string $directory, InputInterface $input, OutputInterface $output)
     {
         $commands = array_filter([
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -444,7 +444,7 @@ protected function installJetstream(string $directory, InputInterface $input, Ou
      *
      * @param  string  $directory
      * @param  \Symfony\Component\Console\Input\InputInterface  $input
-     * @return string
+     * @return array
```
driesvints, line 447: Small return type fix here.

PR #333, opened 2024-04-12
Update `APP_URL` to end in `.localhost`

kemp:
While `.test` is convenient for developers using Laravel Valet, it provides an obstacle for users that don't. PR #240 improved this problem by first checking that the domain resolves correctly. My PR should improve this even further by not requiring DNS modifications for users that aren't using Valet while still allowing them to use `my-app-name.localhost`.

According to the [W3 specification](https://www.w3.org/TR/secure-contexts/#localhost), domains ending in `.localhost` are intended to be treated specially, only resolving as loopback addresses. This means using `.localhost` will allow the browser to treat the site as a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts), allowing developers to use features that would normally require HTTPS (such as the Geolocation API).

This does not break existing features because it only applies to new installations as well as keeps the `canResolveHostname` check with fallback to `http://localhost` in case the DNS resolution fails. In addition, developers using Valet can still use either `.test` or `.localhost`.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -722,7 +722,7 @@ class NewCommand extends Command
      */
     protected function generateAppUrl($name)
     {
-        $hostname = mb_strtolower($name).'.test';
+        $hostname = mb_strtolower($name).'.localhost';
 
         return $this->canResolveHostname($hostname) ? 'http://'.$hostname : 'http://localhost';
     }


kemp:
I'm going to make a few modifications to this PR. One flaw I believe is this will interfere with Valet's site mapping. I don't have a Mac to test at the moment unfortunately.

driesvints:
No plans for this sorry. `.test` is the default in Herd.

kemp:
Thank you for your time. :+1:

I'd still recommend considering falling back to `<app-name>.localhost` for installations that aren't using Valet or Herd (using the DNS check already used) because of the benefits mentioned in the original PR.

jessarcher:
Hey @kemp,

Thank you for teaching me that `*.localhost` is a thing!

Just wondering what are the benefits of `<app-name>.localhost` over just `localhost` for folks not using Herd/Valet? I assume it would be to support running multiple apps at the same time, but you'd still need a central web server on port 80 that could handle the hostname routing.

PR #337, opened 2024-05-09
Only run migration after selecting the database

crynobone:
### Before

`composer.json` from `laravel/laravel` will create `database/database.sqlite` and run `php artisan migrate` before we even select the default database.

image

### After

1. Disable scripts during `composer create-project` to avoid above scenario.
2. Run `composer run post-root-package-install` to copy `.env.example` to `.env`.
3. Run `php artisan key:generate`
4. Select the correct database using prompt.
5. Run migration

https://github.com/laravel/installer/assets/172966/443d4447-8ebc-474f-9f52-2d5adc00cb5d

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -158,9 +158,12 @@ class NewCommand extends Command
         }
 
         $composer = $this->findComposer();
+        $phpBinary = $this->phpBinary();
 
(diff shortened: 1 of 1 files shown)


driesvints (approved):
Yesss 👏

PR #338, opened 2024-05-11
Add `no-migrate` option

milwad-dev:
I've created a project and at first, I wanted to modify the user migration. But when I migrated the migrations, there was no migration to migrate!! and I was forced to run `migrate:rollback` and migrate again and it's bad.
I added a new option for someone who needs to create a project without running the `migrate` command.

```shell
laravel new milwad --no-migrate
laravel new milwad -nm
```


taylorotwell:

jessarcher:
I'd suggest running `php artisan migrate:fresh` in that scenario.

milwad-dev:
Imagine you want to create an app without using a database! So we don't need to migrate automatically. If you need, you can run it as well!

PR #340, opened 2024-05-27
[1.x] Run commands inside the `spin` function.

xiCO2k:
This PR adds the capability to have the `spin` function from Laravel Prompts.

# New vs Old
https://github.com/laravel/installer/assets/823088/53b3c77e-311e-4bfe-8b3b-8c84ffe3e1f6

And of course you can still pass the `-v` and have the exact same output as before.

Cheers.


mpociot:
The spinner can also be used without the pcntl extension installed, as it will render a static spinner in this case: https://github.com/laravel/prompts/blob/main/src/Spinner.php#L50-L52

I would personally prefer it if the message would give a hint at what's currently being installed. Right now it just always says "Installing…" and the user doesn't know what's happening.

Also when some of the commands fail, they fail silently. For example when trying to install NPM dependencies for Breeze with an unsupported Node/NPM version.

xiCO2k:
Yep, I did though about that, but I think we can work on that on a PR after this one.

For the next one I would divide each command into its own process, and for that case we can have better error messages and we can say "npm install", "composer install", etc.

Thanks for the review @mpociot.

taylorotwell:
@xiCO2k awesome - glad you sent this over.

> For the next one, I would divide each command into its own process, and for that case we can have better error messages and we can say "npm install", "composer install", etc.

Can we go ahead and do this before merging?

xiCO2k:
On it!!!

xiCO2k:
@mpociot can you give it a try and let me know what you tinkn?

Thanks.

taylorotwell:
@xiCO2k some conflicts here.

xiCO2k:
ready @taylorotwell

taylorotwell:
Drafting this for a bit while I think about it.

osbre:
Unpopular opinion: seeing the "behind the scenes" of installing dependencies is helpful to get familiar with the ecosystem, and allows them to see the progress or when it gets stuck due to the network.

patriciomartinns:
It would be interesting to have the option to select alternative package managers like pnpm or yarn during the project creation process. Currently, only npm is supported by default, but having the choice could enhance flexibility for users.

PR #341, opened 2024-05-28
Add Ignition

riasvdv:
Since Ignition was [removed from the default project scaffold](https://github.com/laravel/laravel/pull/6406). It would be nice to offer people the option to install it when creating a new project.

Ignition still has many more features than the default error page [now included in Laravel](https://github.com/laravel/framework/pull/51261), like: solutions, direct links to your code editor from the stacktrace, ability to share your exceptions, Livewire context and more.


taylorotwell:
Would prefer people to just Composer install it after the fact. 👍

Issue #342, opened 2024-05-31
NewCommand does not check for herd valet existence


Anticom:
### Installer Version

5.8.1

### Description

https://github.com/laravel/installer/commit/620f1b32e71a7f6bb325da381ca613f626f13160 introduced automatic inference of TLD in case either herd or valet is present on the host machine.
However this method does not check for existence of those tools first, causing the following error to be thrown when generating a new laravel application:

```
In Process.php line 371:

  The command "'herd' 'tld' '-v'" failed.

  Working directory: /path/to/workingdirectory

  Error: proc_open(): posix_spawn() failed: No such file or directory
```

Not only is this bad UX but most improtantly this causes `laravel new` to exit with a non-zero code. This may break scripts created by user automating project scaffolding.

### Steps To Reproduce

1. Make sure neither herd nor valet is on your host system
2. run `laravel new foo`
3. choose default options for promts

driesvints:
cc @xiCO2k: could you maybe send a fix for this one

Anticom:
Additional info:  
I also just noticed NewCommand does not appear to run to completion. I chose to initialize the git repo for me which did not happen.

Anticom:
I re-ran the command with very verbose output, yielding

```
In Process.php line 371:

  [Symfony\Component\Process\Exception\ProcessStartFailedException]
  The command "'herd' 'tld' '-v'" failed.

  Working directory: /path/to/workingdirectory

  Error: proc_open(): posix_spawn() failed: No such file or directory


Exception trace:
  at /Users/anticom/.composer/vendor/symfony/process/Process.php:371
 Symfony\Component\Process\Process->start() at /Users/anticom/.composer/vendor/symfony/process/Process.php:249
 Symfony\Component\Process\Process->run() at /Users/anticom/.composer/vendor/laravel/installer/src/NewCommand.php:766
 Laravel\Installer\Console\NewCommand->getTld() at /Users/anticom/.composer/vendor/laravel/installer/src/NewCommand.php:751
 Laravel\Installer\Console\NewCommand->generateAppUrl() at /Users/anticom/.composer/vendor/laravel/installer/src/NewCommand.php:185
 Laravel\Installer\Console\NewCommand->execute() at /Users/anticom/.composer/vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at /Users/anticom/.composer/vendor/symfony/console/Application.php:1029
[...]

PR #343, opened 2024-05-31
[5.x] Fixes issue if there is no `herd` or `valet` installed

xiCO2k:
This PR fixes the issue #342.

I could not replicate on my machine but based on the error information provided by @Anticom, catching the `ProcessStartFailedException` it should solve.

Cheers.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -14,6 +14,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Process\Exception\ProcessStartFailedException;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
+
 use function Laravel\Prompts\confirm;
 use function Laravel\Prompts\multiselect;
 use function Laravel\Prompts\select;
@@ -787,7 +788,6 @@ class NewCommand extends Command
         return $output !== false ? in_array(dirname($directory), json_decode($output)) : false;
     }
 
-
     /**
      * Runs the given command on "herd" or "valet" cli.
      *
@@ -805,7 +805,8 @@ class NewCommand extends Command
                 if ($process->isSuccessful()) {
                     return trim($process->getOutput());
                 }
-            } catch (ProcessStartFailedException) {}
+            } catch (ProcessStartFailedException) {
+            }
         }
 
         return false;
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -11,9 +11,9 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\Exception\ProcessStartFailedException;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
-
 use function Laravel\Prompts\confirm;
 use function Laravel\Prompts\multiselect;
 use function Laravel\Prompts\select;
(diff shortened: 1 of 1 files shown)


xiCO2k:
@Anticom can you check if this PR solved it?

Thanks.

src/NewCommand.php
```diff
@@ -792,15 +783,29 @@ protected function canResolveHostname($hostname)
...
-                $output = json_decode(trim($process->getOutput()));
+            try {
+                $process->run();
 
-                return in_array(dirname($directory), $output);
+                if ($process->isSuccessful()) {
+                    return trim($process->getOutput());
+                }
+            } catch (ProcessStartFailedException) {
             }
```
Anticom, line 809: Technically this is returning `null` and not false in case neither `herd` nor `valet` are present, isn't it?

In `isParked´ you're doing a strict compare with `false`, which might be an issue in such case.

xiCO2k: I don't think it will return `null`. If herd or valet is present I except the output to be correct.

I can increase the checking on the `json_decode` but not sure if is needed.

What do you think @driesvints?

driesvints: @xiCO2k but what if neither is present like @Anticom says. What will happen then?

xiCO2k: that will throw the `ProcessStartFailedException` and we will ignore it and try `valet` is the same happens we will ignore and then after the loop we will return `false` as before.

driesvints (approved):

src/NewCommand.php
Anticom: Sorry guys, I missed it since it wasn't shown in the diff. My bad!

Anticom:
> @Anticom can you check if this PR solved it?
> 
> Thanks.

I'm getting promted for a password now multiple times? Is this normal in 5.x branch of installer?

First was after setting app key:
```
   INFO  Application key set successfully.
```
second time after:

```
   INFO  Application ready in [foo]. You can start your local development using:
```

All I did to check out your MR was cloning your fork into ~/.composer/vendor/laravel/installer and checking out your branch.

xiCO2k:
> > @Anticom can you check if this PR solved it?
> > Thanks.
> 
> I'm getting promted for a password now multiple times? Is this normal in 5.x branch of installer?
> 
> First was after setting app key:
> 
> ```
>    INFO  Application key set successfully.
> ```
> 
> second time after:
> 
> ```
>    INFO  Application ready in [foo]. You can start your local development using:
> ```
> 
> All I did to check out your MR was cloning your fork into ~/.composer/vendor/laravel/installer and checking out your branch.

Nice! the password prompts does not seem related with this.

Thanks for checking it out.

Anticom:
Yeah I figured it probably has nothing to do with the changes you made to fix the issue. I'll keep you posted once the fix is released.

@xiCO2k 
Is there any chance we can get a 5.8.2 today? 🙂

rjindael:
Would love if this could be pushed to a new version 5.8.2, this currently breaks the Laravel installer entirely on WSL2 :/

driesvints:
I just published a new release for this one.

Anticom:
I updated using composer and am still getting promted for password.

xiCO2k:
Ok but that worked without any error?

Anticom:
@xiCO2k well if I enter the password correctly, it works as expected w/o any errors.

Maybe @rjindael can dis-/confirm the issue with the password promts?

rjindael:
> @xiCO2k well if I enter the password correctly, it works as expected w/o any errors.
> 
> Maybe @rjindael can dis-/confirm the issue with the password promts?

I think my original reply was worded incorrectly , what I had meant was that this PR *did* solve #342 and that it should be merged as the issue described in #342 currently causes the installer on WSL2 to not work at all

rjindael:
> @rjindael can you create an issue with the error you are getting?
> 
> Thanks.

Sorry, I think I was not clear, I am not experiencing any issues now, only previously did before this PR was merged. 🙂

Anticom:
@rjindael So you're not getting promted for passwords during execution of `laravel new ...` ?

rjindael:
> @rjindael So you're not getting promted for passwords during execution of `laravel new ...` ?

No

PR #206, opened 2021-08-03
[4.x] added support for custom template

myckhel:
These changes enables users to specify an optional custom template for their new laravel project.

Unlike other framework's installers such as `react-native-init` accepts optional custom template option to be used.

## Usage
Created an example template which the welcome page title was changed from `Laravel` to `Laravel Example Template`
```bash
laravel new blog --template myckhel/laravel-template-example
```

## Prerequisites
- Template repository should have been submited to composer's [packagist](https://packagist.org)
- Template repository should have at least a release tag for it to be installed.
- Template repository should have a `master` branch if the `--dev` installer option is specified as installer expects.

PR #348, opened 2024-06-19
[5.x] Add support for custom template

joshmanders:
This feature was [attempted 3 years ago](https://github.com/laravel/installer/pull/206) but was rejected. But with changes in the landscape and more people finding out about the awesome benefits of Laravel coming from JavaScript world, they're used to easy to get started with `create-*-app` style installers.

The default laravel installer is great, but as end users there's a lot left to be desired. Being able to quickly scaffold out a jetstream or a breeze app is fine, but allowing the community to build out more robust starting skeletons and giving a Laravel sanctioned way to install those skeletons will improve people picking up Laravel a lot quicker.

## Usage
You can now just utilize the `--template` flag to specify a non-default skeleton template to use, (this defaults to `laravel/laravel`) like so:

```sh
laravel new blog --template foo/laravel-blog
```

## Prerequisites
* Template repository should have been submited to composer's [packagist](https://packagist.org)
* Template repository should have at least a release tag for it to be installed.
* Template repository should have a `master` branch if the `--dev` installer option is specified as installer expects.

## Considerations
Some of the options should be ignored when using a template that isn't the default, so maybe I can amend this PR to disable breeze/jetstream/etc if `--template` flag is used.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -61,7 +61,8 @@ class NewCommand extends Command
             ->addOption('phpunit', null, InputOption::VALUE_NONE, 'Installs the PHPUnit testing framework')
             ->addOption('prompt-breeze', null, InputOption::VALUE_NONE, 'Issues a prompt to determine if Breeze should be installed (Deprecated)')
             ->addOption('prompt-jetstream', null, InputOption::VALUE_NONE, 'Issues a prompt to determine if Jetstream should be installed (Deprecated)')
-            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists');
+            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Forces install even if the directory already exists')
+            ->addOption('template', 't', InputOption::VALUE_OPTIONAL, 'The laravel template to use', 'laravel/laravel');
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's better to just use `composer create-project` for this purpose. Otherwise the `new` command has a bunch of options that don't even make sense.

joshmanders:
> I think it's better to just use `composer create-project` for this purpose. Otherwise the `new` command has a bunch of options that don't even make sense.

That's why I had the considerations section. But there's also a bunch of things the `new` command does that `composer create-project` doesn't do that adds a bunch of extra overhead for newbies coming from other places.

NickSdot:
Good. Groundwork for an official Laravel marketplace. 

People can distribute their packages how they want and without limitations.

Then there is the marketplace. If paid, or freemium, packages want to get listed on the Marketplace to benefit from the Laravel effect they pay. 

Win-win-win. Go Taylor do that already.

joshmanders:
I have forked the installer to create a `laranew` global CLI for this to hopefully show Taylor that the community will benefit from this ability and we can think of bringing it in natively. https://github.com/aniftyco/laranew

PR #353, opened 2024-08-14
Add database credentials support

CyberPunkCodes:
The current installation process is a bit clunky. You have to create a database matching the Laravel application, it assumes your username should be root, and with no password.

This PR asks you during the CLI installation process for the database name, user, and password. Leaving blank will use the default values just as Laravel always has used.

It also supports passing the options to the command line via: `dbname`, `dbuser`, and `dbpass`

Screen Shot 2024-08-14 at 3 06 07 PM

---

I don't have all the databases setup. Is there a use case, ie: using one of the other database options other than MySQL, that we wouldn't want (or need) to ask the credential questions?

I would like to get some feedback to ensure this works how people would like it to work for the various database types.


taylorotwell:

CyberPunkCodes:
@taylorotwell The current install process, at least for MySQL, is clunky. It assumes you want the database named the exact same as the project, it does not allow you to change the user from `root`, and it does not support a password. You can't run the migrations seamlessly through the installer process. You have to switch over to your editor and edit the `.env` file before running the migrations in the installer.

You can see some of the issue discussed here: [https://laracasts.com/index.php/discuss/channels/laravel/how-to-create-a-new-laravel-project-when-the-root-password-is-not-empty](https://laracasts.com/index.php/discuss/channels/laravel/how-to-create-a-new-laravel-project-when-the-root-password-is-not-empty)

> we need to be very careful regarding the amount of code we include

It's two small sections. One for `dbuser` and another for `dbpass`.

This is the last time I waste my time with PRs for anything laravel.

snapey:
How do you create a package to be included in the framework install process?

antimech:
Personally, I'm fine with editing the `.env` file just before answering "yes" to the question "Would you like to run the default database migrations?".

CyberPunkCodes:
> Personally, I'm fine with editing the `.env` file just before answering "yes" to the question "Would you like to run the default database migrations?".

Then what is the point of using a handy dandy installer? If you can't do what you need in the installer, then the installer is falling short.

Issue #352, opened 2024-08-13
Running with `--no-interaction` still prompts about SQLite file


jacksleight:
### Installer Version

5.8.3

### Description

When running the installer with the `--no-interaction` option you're still prompted about creating the SQLite database file:

```
laravel new example-app --no-interaction

...


   WARN  The SQLite database configured for this application does not exist: database/database.sqlite.

 ┌ Would you like to create it? ──────────────────────┐
 │ ● Yes / ○ No                                       │
 └────────────────────────────────────────────────────┘
```

I thought adding the `--database=sqlite` option might work, but it doesn't seem to.

### Steps To Reproduce

Run:

```
laravel new example-app --no-interaction
```

PR #354, opened 2024-08-20
Fix `--no-interaction` SQLite file prompt

jacksleight:
Fixes https://github.com/laravel/installer/issues/352

This prompt comes from the `migrate` command, and [by the looks of that](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Console/Migrations/MigrateCommand.php#L196) the only way to ensure the SQLite file is created without a prompt is to add `--force`. Adding `--no-interaction` to `migrate` results in no file being created, which then errors when the migrations run.

Hopefully this is OK but let me know if you'd like any changes.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -211,7 +211,7 @@ class NewCommand extends Command
                     $commands = [
                         trim(sprintf(
                             $this->phpBinary().' artisan migrate %s',
-                            $input->getOption('no-interaction') ? '--force' : '',
+                            $database === 'sqlite' && ! $input->isInteractive() ? '--force' : '',
                         )),
                     ];
                     $this->runCommands($commands, $input, $output, workingPath: $directory);
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -214,7 +214,7 @@ class NewCommand extends Command
                     $commands = [
                         trim(sprintf(
                             $this->phpBinary().' artisan migrate %s',
-                            ! $input->isInteractive() ? '--force' : '',
+                            ! $input->isInteractive() ? '--no-interaction' : '',
                         )),
                     ];
                     $this->runCommands($commands, $input, $output, workingPath: $directory);
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -211,12 +211,14 @@ class NewCommand extends Command
                     if ($database === 'sqlite') {
                         touch($directory.'/database/database.sqlite');
                     }
+
                     $commands = [
                         trim(sprintf(
                             $this->phpBinary().' artisan migrate %s',
                             ! $input->isInteractive() ? '--no-interaction' : '',
                         )),
                     ];
+
                     $this->runCommands($commands, $input, $output, workingPath: $directory);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -208,9 +208,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 $this->configureDefaultDatabaseConnection($directory, $database, $name);
 
                 if ($migrate) {
-                    $this->runCommands([
-                        $this->phpBinary().' artisan migrate',
-                    ], $input, $output, workingPath: $directory);
+                    $commands = [
+                        trim(sprintf(
+                            $this->phpBinary().' artisan migrate %s',
+                            $database === 'sqlite' && ! $input->isInteractive() ? '--force' : '',
```
crynobone, line 214: Wouldn't this also can happen on MySQL database as well?

crynobone: Or, shouldn't it be executed with `--no-interaction` as well

jacksleight: > Wouldn't this also can happen on MySQL database as well?

Ah yeah probably, I will update.

> Or, shouldn't it be executed with --no-interaction as well

I thought so, but `--no-interaction` prevents the database being created, see https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Console/Migrations/MigrateCommand.php#L200, which just results in an error when the migrations then run. Not exactly sure why the `migrate` command works like that but it does.

crynobone: I believe `--no-interaction` behaviour on migrate should be respected here if you initiate laravel new without interaction.

jacksleight: OK so call `php artisan --force --no-interaction` yeah?

crynobone: I wouldn't recommend `--force` unless `new` command also has the option, this would make it hard to maintain if in future `--force` contains action that not inline with `--no-interaction`.

jacksleight: OK but if you just do `php artisan migrate --no-interaction` the migrate command will throw an exception:

> Illuminate\Database\QueryException
> Database file at path [/.../database/database.sqlite] does not exist.

crynobone: Why don't just touch the file within this command instead? Would that make more sense?

jacksleight: Yeah can do. That would work.

crynobone (approved):

stancl:
Not sure if I'm doing something wrong, but even on `laravel/installer:dev-master` I'm getting
> The SQLite database configured for this application does not exist: database/database.sqlite.

with:

`laravel new --no-interaction --git --jet --stack=livewire --api --verification --pest --database=sqlite test`

Also, `--force` having an effect here seems wrong since the `--help` comment for `--force` is:
> Forces install even if the directory already exists

So that behavior should all be under `--no-interaction`.

Either way, on the latest `dev-master`, the command above prompts me for both creating the DB and re-running migrations, and `--force` doesn't change anything 🤷🏻‍♂️ 

Confirmed with `composer global show laravel/installer` that I'm using the right commit:

```
source   : [git] https://github.com/laravel/installer.git db5f79dda5bd9b84c0777da435d2147c5caa4068
```

PR #363, opened 2024-11-08
Undo git question comment

RobertBoes:
Recently noticed the git question was removed, it's quite a useful options to have.

Perhaps slightly off-topic; the changes were made here https://github.com/laravel/installer/commit/667c610382f2052e69ff60b10811fdaa9376aeef without a PR and not a mention in the release notes. Shouldn't changes like these at least be mentioned in the release notes? That way it would also likely include reasoning as to why it was removed, now it's just commented out.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -142,9 +142,9 @@ class NewCommand extends Command
             ) === 'Pest');
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
We ask so many questions on install. I don't think we should do this.

PR #368, opened 2024-12-06
Remove any trailing slash from application name

jasonmccreary:
This PR addresses two, minor issues.

First, due to a logic bug when testing for Herd or Valet, a default `.test` extension could never be reached. This ultimately leads to `APP_URL` always being `http://localhost` instead of a potentially valid `http://app-name.test`.

Second, if a developer appends a trailing slash to the name everything is installed and the application runs. However, all tests will fail until the developer realizes the `APP_URL` was set to `http://app-name/.test`. While an arguably an incorrect use case, it's a pretty easy fix to avoid such an annoying experience.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -810,7 +810,7 @@ class NewCommand extends Command
      */
     protected function getTld()
     {
-        return $this->runOnValetOrHerd('tld') ?? 'test';
+        return $this->runOnValetOrHerd('tld') ?: 'test';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jasonmccreary:
This test case may not be possible in GitHub actions as their environment may have different responses for `gethostbyname`. It runs locally on my Mac as it can always lookup a local domain.

I can remove this test case, but will leave it in case the reviewer wishes to run it locally to verify my change.

src/NewCommand.php
```diff
@@ -159,7 +159,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $this->validateDatabaseOption($input);
         $this->validateStackOption($input);
 
-        $name = $input->getArgument('name');
+        $name = mb_rtrim($input->getArgument('name'), '/\\');
```
didix16, line 162: This change breaks the minimum required PHP version. This function is introduced on PHP 8.4

PR #376, opened 2024-12-11
Ensure Symfony/mbstring-polyfill version 1.31 or higher

### LINKED ISSUE #375: `laravel new` crashes if system has PHP version 8.3.* or below and `Symfony/polyfill-mbstring` version 1.30 or below
### Installer Version

5.11

### Description

This binary uses [`mb_rtrim`](https://github.com/laravel/installer/blob/09627d33c1e367844f36cf5ed11b3be101d7ad14/src/NewCommand.php#L196) which was only added to the `Symfony/mbstring-polyfill` in version [1.31](https://github.com/symfony/polyfill-mbstring/commit/74b87a0c69bc80c8b7bd7cf630dea24f4573f0e3). If you installed Laravel before version 1.31 (such as version 1.30) which was released in Sept 9, you would likely have `Symfony/mbstring-polyfill` version 1.30 or lower.
https://github.com/symfony/polyfill-mbstring/releases/tag/v1.31.0

`composer.json` does not specify a version of `mbstring-polyfill`, so it's a soft dependency on version 1.31 without a guarantee. Other symfony dependencies specify version `"^1"`, so 1.30 can sneak past without being updated.

### Steps To Reproduce

__Ensure that you are on a PHP version before 8.4, such as 8.3.* where before `mb_rtrim` is added to the `mbstring` extension or else disable the `mbstring` extension to use the polyfill.__

##### Global method
Run
```sh
composer global require symfony/polyfill-mbstring 1.30.0
```
to simulate having installed Laravel before Sept 9.

Then run
```sh
composer global require laravel/installer 5.11
```

Now run `laravel new my_app`, hit enter twice to select the testing framework and you should see:
```
PHP Fatal error:  Uncaught Error: Call to undefined function Laravel\Installer\Console\mb_rtrim() in <your path to .composer/vendor>/laravel/installer/src/NewCommand.php:162
```
- Note: You may not see this error on Windows, and the script may fail silently. For me on Fedora 41 I can see the exception. I don't understand enough about the Laravel ecosystem to know why this happens.

#### Local method

Clone the repository and `cd` to it.
- To simulate the bad state of dependencies run
  ```sh
  composer require symfony/polyfill-mbstring 1.30.0
  ```
- Run `composer install`
- Run `./bin/laravel new my_app`
- Hit enter twice to select the testing framework
- You should see
  ```
  PHP Fatal error:  Uncaught Error: Call to undefined function
  Laravel\Installer\Console\mb_rtrim() in <your path to the project>/src
  NewCommand.php:162
  ```
  - Note: You may not see this error on Windows, and the script may fail silently. For me on Fedora 41 I can see the exception. I don't understand enough about the Laravel ecosystem to know why this happens.

ellnix:
Fixes #375

Please let me know if I should include further details, this is my first time contributing to anything in the Laravel ecosystem :smile:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "illuminate/support": "^10.20|^11.0",
         "laravel/prompts": "^0.1.18|^0.2.0|^0.3.0",
         "symfony/console": "^6.2|^7.0",
-        "symfony/process": "^6.2|^7.0"
+        "symfony/process": "^6.2|^7.0",
+        "symfony/polyfill-mbstring": "^1.31"
     },
     "require-dev": {
         "phpstan/phpstan": "^1.10",


crynobone:
But Laravel Framework itself requires `mbstring` extension

ellnix:
> But Laravel Framework itself requires `mbstring` extension

There are changes to the mbstring extension on PHP 8.4 that the polyfill backports. Besides, this polyfill is already a dependency of this project through [`Symfony/console`](https://github.com/symfony/console/blob/fefcc18c0f5d0efe3ab3152f15857298868dc2c3/composer.json#L20).

crynobone (approved):

hackel:
@ellnix This didn't fix the issue for me.  Composer refused to install v5.11.1 of the installer because I already had an earlier symfony/polyfill-mbstring installed.  I still had to manually upgrade symfony/polyfill-mbstring, then I could upgrade and run laravel/installer successfully.

ellnix:
> @ellnix This didn't fix the issue for me. Composer refused to install v5.11.1 of the installer because I already had an earlier symfony/polyfill-mbstring installed. I still had to manually upgrade symfony/polyfill-mbstring, then I could upgrade and run laravel/installer successfully.

That is intended behavior as far as I understand. I don't know of a way for a package to forcibly update a dependency that you have locked to an earlier version. With this change, you are prompted to install a compatible version instead of the command failing with a strange error message (or none at all) when you try to run `laravel new` on PHP 8.3 and earlier.

Please let me know if I'm wrong, like I said I'm new to the Laravel ecosystem, I'm going by knowledge of other dependency management systems.

PR #378, opened 2024-12-12
Fix check for current directory

### LINKED ISSUE #377: Cannot create a project in current directory (".")
### Installer Version

5.11

### Description

When the script checks if the directory exists, there is a condition to add an exception for the current directory, but it does a string comparison between "." and `getcwd()`. Therefore, that exception never runs. `-f` cannot be run with ".", so there is no way to actually create a project in the current directory.

### Steps To Reproduce

- `mkdir my_app`
- `cd my_app`
- `laravel new .`
  ```
  In NewCommand.php line 792:
                               
  Application already exists!  
  ```

ellnix:
Fix #377

This was the smallest change I could make to fix the issue. It might be better if instead of comparing to `"."` we set change `getInstallationDirectory()` from
```php
        return $name !== '.' ? getcwd().'/'.$name : '.';
```
to
```php
        return $name !== '.' ? getcwd().'/'.$name : getcwd();
```

Any feedback would be appreciated.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -822,7 +822,7 @@ class NewCommand extends Command
      */
     protected function verifyApplicationDoesntExist($directory)
     {
-        if ((is_dir($directory) || is_file($directory)) && $directory != getcwd()) {
+        if ((is_dir($directory) || is_file($directory)) && $directory != getcwd() && $directory != '.') {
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to mess with this right now.

Issue #250, opened 2023-01-10
Define version in new project


kromodoro:
I'm starting in Laravel this week, and I noticed that there are 3 ways to create a new project.

But only via composer I can define the framework version.

Via CURL or Laravel installer I don't have any way to set the version I want to start a new project.

![image](https://user-images.githubusercontent.com/21073299/211618210-036b1b9d-91b6-4260-a9a3-aa19e1993cc0.png)
> Laravel Installer 4.2.17

driesvints:
You always get the latest version which should be the default for every new project 👍

Issue #229, opened 2022-06-10
Set version to install


perepaul:
The laravel install is very robust as it helps with many thing that would be done manually, ie. setting up a github repo, initialising git and lots more. I have been receiving requests from employers to run laravel 8 or versions below 8, please can you add a flag were we can set the version of laravel we'd want to install?

laravel new example-app --version ^8.0

it will be really helpful for legacy support.

driesvints:
We have no plans to allow people to install older Laravel versions. Please only start new projects with the latest Laravel version.

srmklive:
@driesvints are there still plans to add support for this. I actually have a PR ready for this.

driesvints:
No sorry. Only latest Laravel is supported and we’ll probably never change that. We don’t want to recommend installing older versions.

PR #383, opened 2025-01-03
Add `--target` option to specify Laravel version to install

miclaus:
I've added a `--target` option to allow developers to specify which version of Laravel they want to install. This feature has been frequently requested over the years (see #229 in 2022, #250 in 2023, #229 again in 2023, and https://github.com/laravel/installer/pull/319#issuecomment-2001962089 in 2024).

Recently, I personally encountered the need for this feature when working with a client's server that runs multiple PHP applications that are dependent on PHP 8.1. The version of PHP cannot be changed, and the client is not willing to pay for another server, making it necessary to install Laravel 10 rather than the latest version.

What's the alternative? The documentation doesn't mention any (anymore). As of Laravel 11, documentation on installing with `composer create-project` was removed. Docs is pointing to the laravel installer, which everyone I believe, myself included, enjoys using.

This PR would allow:
```bash
laravel new example-app --target=^10.0
```

<details>
<summary>Also consider "Docker Installation using Sail"</summary>

The `laravel.build` (sail-server) is also affected by the absence of this feature, since it relies on the Laravel installer behind the scenes. To understand how the script creates a new project with an older Laravel version, you'd need to inspect not just the `laravel.build` script, but also the Laravel installer's code. This adds unnecessary complexity.

"Docker Installation using Sail" is recommended like so:
```bash
curl -s "https://laravel.build/example-app" | bash
```
Older Laravel docs (e.g. see [Docker Installation Using Sail](https://laravel.com/docs/10.x/installation#docker-installation-using-sail)) don't mention that this command installs the latest Laravel version. **This leads to confusion (or is even left unnoticed) when developers expect to install the version they are currently reading about in the documentation.**

I've already prepared a solution for myself (see changes in https://github.com/miclaus/sail-server/pull/2 and https://github.com/miclaus/sail-server/pull/3) for the sail-server that adds a `version` query parameter for passing the `--target` option to the Laravel installer. This change could easily be added for older docs. Together with this PR's feature, the following would be possible for e.g. Laravel 10:
```bash
curl -s "https://laravel.build/example-app?version=10" | bash
```
</details>

---

This PR adds flexibility without altering the default experience, and would be particularly developer-friendly for those working with legacy systems or in resource-constrained environments.

I totally agree that installing the current Laravel version should be recommended. Still, IRL sometimes we need to work with what we have, and this change would simplify the process a lot.

Please consider this. I'm sure we can figure something out.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -42,7 +42,8 @@ class NewCommand extends Command
             ->setName('new')
             ->setDescription('Create a new Laravel application')
             ->addArgument('name', InputArgument::REQUIRED)
-            ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release')
+            ->addOption('dev', null, InputOption::VALUE_NONE, 'Installs the latest "development" release. Cannot be used with --target')
+            ->addOption('target', null, InputOption::VALUE_OPTIONAL, 'Installs the specified version of Laravel. Cannot be used with --dev')
             ->addOption('git', null, InputOption::VALUE_NONE, 'Initialize a Git repository')
             ->addOption('branch', null, InputOption::VALUE_REQUIRED, 'The branch that should be created for a new repository', $this->defaultBranch())
             ->addOption('github', null, InputOption::VALUE_OPTIONAL, 'Create a new repository on GitHub', false)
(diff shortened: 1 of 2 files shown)


taylorotwell:

miclaus:
@taylorotwell Thank you for taking the time to review my PR!

Honestly, this change would significantly improve the quality of life for Laravel developers, especially those working in constrained environments.

It's a feature frequently requested by the community and introduces a just tiny portion of simple logic that actually requires one less test to maintain.

Could you please let me know if there are specific concerns or further adjustments I could make to align this PR with the framework's goals? I'd be happy to help.

PR #385, opened 2025-01-11
Fix Laravel installer on Windows WSL

### LINKED ISSUE #384: Laravel installer fails on PHP 8.3 using Windows WSL
### Installer Version

5.11.1

### Description

Attempting to run `laravel new` on WSL when using PHP 8.3 leads to the following error:
```
PHP Fatal error:  Uncaught Error: Call to undefined function Laravel\Installer\Console\mb_rtrim() in /home/jaytame/.config/composer/vendor/laravel/installer/src/NewCommand.php:162
```

I have tested on MacOS and it works fine on PHP 8.3 there, it seems like `mb_trim` gets polyfilled on Mac.  

### Steps To Reproduce

- Run WSL on Windows
- Install a version of PHP older than 8.4
- Run `laravel new`

JasonTame:
Resolves #384 

The missing `mb_trim()` function causes `laravel new` to crash on Windows WSL when running a PHP version lower than 8.4.

Considering the command isn't being used to strip multi-byte characters, `rtrim` should work fine and won't lead to any issues for users who are not yet on PHP 8.4.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -193,7 +193,7 @@ class NewCommand extends Command
         $this->validateDatabaseOption($input);
         $this->validateStackOption($input);
 
-        $name = mb_rtrim($input->getArgument('name'), '/\\');
+        $name = rtrim($input->getArgument('name'), '/\\');
 
         $directory = $this->getInstallationDirectory($name);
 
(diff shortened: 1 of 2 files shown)


crynobone:
This PR #376 should already fixed the issue for PHP 8.3 and below.

JasonTame:
@crynobone It didn't fix it for me on Windows WSL though, the installation crashes with an exception even though I had the `mbstring` extension. 

If the consensus is that this is user error on my side, that's fine, it was easy enough to resolve anyway - I just would hate for new Laravel devs on Windows to give up when the installer crashes for them without a clear indication of why.

PR #400, opened 2025-02-25
Use pest command in github workflow when using pest

laserhybiz:


taylorotwell:
This should only happen when using a starter kit.

Issue #329, opened 2024-03-31
Database.sqlite is created even if you slect mysql


rabol:
### Installer Version

5.7.0

### Description

Even if MySql is selected as database, the database.sqlite is created

### Steps To Reproduce

laravel new xxx
then in the section of database select MySQL, say no to migrate and look at the database directory and you will see that there is a database.sqlite

crynobone:
This seems to be the expected behavior with `compsoer install` used during the installer + post create/install command. Feel free to attempt PR changes to the behavior.

PR #405, opened 2025-02-28
Remove sqlite from composer.json when not selected as database

MrYamous:
If we choose anything else than sqlite as a database, composer post create script is still looking for/creating a sqlite file.
This PR aims to change to behavior when sqlite is not selected

Already reported in #329


taylorotwell:
Not sure this is really a big issue.

Issue #404, opened 2025-02-27
Laravel 12 : The database= option doesn't seem to work as expected


happytodev:
### Installer Version

5.12.0

### Description

When I launch the following : 

`laravel new l12pgsql --database=pgsql --livewire`

or the following

`laravel new l12mysql --database=mysql --livewire`

The Laravel installer installs sqlite.

```
DB_CONNECTION=sqlite
# DB_HOST=127.0.0.1
# DB_PORT=3306
# DB_DATABASE=laravel
# DB_USERNAME=root
# DB_PASSWORD=

```

Maybe I'm wrong, but it sounds like a bug for me.

### Steps To Reproduce

Just install a new laravel app with following command : 

`laravel new l12pgsql --database=pgsql --livewire`

or the following

`laravel new l12mysql --database=mysql --livewire`

MrYamous:
Looks like this is on purpose https://github.com/laravel/installer/blob/61db11d1f05d5496d0ed8a34fca792bdc16edec0/src/NewCommand.php#L507-L512

But this is quite unexpected behaviour

PR #406, opened 2025-02-28
[12.x] Restore the ability to choose a database engine during the installation procedure

happytodev:
This PR brings a fix to the #404 issue.

Maybe I'm wrong, but the line removed restores the possibility to choose a database engine during installation procedure.

Tell me if something is missing.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -507,8 +507,6 @@ class NewCommand extends Command
         if ($this->usingStarterKit($input)) {
             // Starter kits will already be migrated in post composer create-project command...
             $migrate = false;
-
-            $input->setOption('database', 'sqlite');
         }
 
         if (! $input->getOption('database') && $input->isInteractive()) {


MrYamous:
I dont know if forcing migrate to false should stay too. Regarding to comment, I think this is related to sqlite only

taylorotwell:
I personally want things to be as quick as possible for starter kits so prefer to just ship it with SQLite as the default so it's ready to go with no more choices.

happytodev:
> I personally want things to be as quick as possible for starter kits so prefer to just ship it with SQLite as the default so it's ready to go with no more choices.

I fully understand your point of view, but I think user first and personally I prefer to have the choice of shipping fast by choosing the right database engine directly during installation instead of changing it after installation.

But you're the boss, boss 😉

MrYamous:
The choice is given but ignored. I still believe this behavior is misleading.
Let's remove the database option then ?

PR #412, opened 2025-03-12
Adding PHPUnit option back when no starter kit is selected

tnylea:
This PR will add back the functionality that allows a user to select **Pest** when not using a starter kit.

This was originally implemented, but it got removed somewhere along the way.


tnylea:
Replacing this PR with the following PR: https://github.com/laravel/installer/pull/413

PR #413, opened 2025-03-12
Adding Pest option back in Laravel app and improve pest experience

tnylea:
This PR will add the `pest` option back to the Laravel installer and run drift in a default Laravel app to convert the PHPunit class-based tests to Pest.

@nunomaduro and I collaborated to create a version that would not require the starter kits to need the `Pest.php` file in the test directory but will init that file and make necessary modifications if the user selects `pest` as their testing option.

I'll do some more testing, and then I'll move this out of Draft mode.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -6,6 +6,8 @@ use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Composer;
 use Illuminate\Support\ProcessUtils;
 use Illuminate\Support\Str;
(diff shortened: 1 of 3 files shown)


Jubeki:
Hey @tnylea thanks for all the good work.

Maybe it would be a great idea to add a notice if a new version of the installer is available, because many developers do not upgrade their global composer dependencies regularly.

You can find some ideas here, though there would be considerations needed for Herd support:
https://github.com/laravel/installer/pull/308

PR #415, opened 2025-03-31
Only do extra things in official starter kits / default laravel

### LINKED ISSUE #410: `--useing` doing too much
### Installer Version

5.14

### Description

When using a userland starter using the `--using` flag, the installer is doing too much. If no `post-root-package-install` is set the installer just flat out kills and doesn't finish any steps (not all starters may need this script), and if the starter doesn't necessarily need other things such as running npm install and build, it still prompts for and runs those.

Those extra steps should only happen for the official Laravel starters. Otherwise you get unintended side effects.

For example, I use Skeletor in all my starters and since Skeletor prompts the user for additional information specific to that starter, the laravel installer clobbers that and outputs similar information and asks things that shouldn't be asked. 

Image

As you can see from this screenshot, the `> @php -r "file_exists('.env') || copy('.env.example', '.env');"` is the beginning of the additional unnecessary steps performed by the installer.

I can submit a PR fixing this if necessary. 

### Steps To Reproduce

```
laravel new --using starter/laravel my-app
```

joshmanders:
Closes #410 

With the addition of `--using` flag, we now have the ability to create userland starter kits, but unfortunately if your starter kit deviates just a little bit away from how Laravel officially does things, such as maybe the starter kit doesn't use a database at all, just a static file system, getting asked about what database you use and running commands to alter the starter based on that could be problematic. This PR only executes these prompts if you're in an official laravel starter kit or the default laravel skeleton.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -151,14 +151,12 @@ class NewCommand extends Command
             }
         }
 
-        if ($this->usingLaravelStarterKit($input) || ! $input->getOption('using')) {
-            if (! $input->getOption('phpunit') && ! $input->getOption('pest')) {
-                $input->setOption('pest', select(
-                    label: 'Which testing framework do you prefer?',
-                    options: ['Pest', 'PHPUnit'],
-                    default: 'Pest',
-                ) === 'Pest');
-            }
+        if (! $input->getOption('phpunit') && ! $input->getOption('pest') && ! $input->getOption('using')) {
+            $input->setOption('pest', select(
+                label: 'Which testing framework do you prefer?',
+                options: ['Pest', 'PHPUnit'],
+                default: 'Pest',
+            ) === 'Pest');
         }
     }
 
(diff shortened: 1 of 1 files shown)


devajmeireles:
Wouldn't a better way to fix this be to check if `composer.json` has `post-root-package-install` before running it, rather than just running it on the official Laravel starter kits?

joshmanders:
> Wouldn't a better way to fix this be to check if `composer.json` has `post-root-package-install` before running it, rather than just running it on the official Laravel starter kits?

It's doing more than just executing that command. Asking for a testing suite choice, copying `.env.example` to `.env` and updating the `.env` file could also be issues for custom starter kits.

joshmanders:
@taylorotwell 
> I think we would need to think about this one a bit. The env stuff especially feels like something you would want to keep.

That's a redundant extra step in the installer anyway as the official skeleton and all the starter kits already do that in the `post-root-package-install`

PR #424, opened 2025-06-21
feat: add support for multiple package manager

pushpak1300:
This PR adds support for additional popular Node.js package managers Yarn, Bun, and PNPM. Currently, the Laravel installer only supports npm, which can be restrictive for third-party starter kits that use a different package manager. Supporting these out of the box would make the installer more flexible and developer-friendly.

This also introduces a new --node option, allowing users to default to detected package manager. Passing `--npm` will result into using npm package manger.

This is a non-breaking change.

Thanks.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -834,14 +834,18 @@ class NewCommand extends Command
     protected function getPackageManagerCommands(string $directory, ?string $packageManager = null): array
     {
         if (! $packageManager) {
-            if (file_exists($directory.'/bun.lockb')) {
-                $packageManager = 'bun';
-            } elseif (file_exists($directory.'/pnpm-lock.yaml')) {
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -303,23 +304,32 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 $output->writeln('');
             }
 
-            $runNpm = $input->getOption('npm');
+            $runNpm = $input->getOption('npm') || $input->getOption('node');
```
MrPunyapal, line 307: Js syntax 😂

MrPunyapal, line 307: Maybe?
```suggestion
            $runNpm = $input->getOption('npm') ?: $input->getOption('node');
```

taylorotwell:

PR #426, opened 2025-07-06
List possible values for database driver

dmarcoux:
Nobody has to guess anymore what the possible values are for the --database option.


src/NewCommand.php
```diff
@@ -561,8 +563,8 @@ protected function databaseOptions(): array
      */
     protected function validateDatabaseOption(InputInterface $input)
     {
-        if ($input->getOption('database') && ! in_array($input->getOption('database'), $drivers = ['mysql', 'mariadb', 'pgsql', 'sqlite', 'sqlsrv'])) {
-            throw new \InvalidArgumentException("Invalid database driver [{$input->getOption('database')}]. Valid options are: ".implode(', ', $drivers).'.');
+        if ($input->getOption('database') && ! in_array($input->getOption('database'), self::DATABASE_DRIVERS)) {
+            throw new \InvalidArgumentException("Invalid database driver [{$input->getOption('database')}]. Possible values are: ".implode(', ', self::DATABASE_DRIVERS).'.');
```
dmarcoux, line 567: We are validating the --database option here, but what the users provide is a value for that option. This is why I've changed the text a bit to better reflect this.

PR #431, opened 2025-08-03
Fix database starter kits

mathiasgrimm:
**Previously:** Always using sqlite.
    
**With this PR:** Prompting for database when Interactive.


taylorotwell:
Let me think on this one. I kinda like that you don't have to make too many decisions and everything "just works".

PR #434, opened 2025-08-13
No Auth commands not new skeleton repos

AhmedAlaa4611:
Description
---
I don’t think this is the best approach. We’ve introduced three additional repositories, and we’ll have to maintain them every time we update the main `laravel/laravel` skeleton. Since we already have the `php artisan install:api` command, it would make sense to also have commands like `php artisan install:react`, `php artisan install:vue`, and `php artisan install:livewire`. These commands should update only the necessary files, rather than requiring the creation and maintenance of three separate repositories.


taylorotwell:
I would probably be OK with that if someone wants to work on a PR to make it happen. 👍

PR #438, opened 2025-09-29
Check for new version of installer on `new`

joetannenbaum:
This PR checks for any newer versions of the installer on Packagist and offers to self-update for the user if one exists. 

It only checks Packagist at most once per 24 hours, and sends the `If-Modified-Since` header for maximum caching.

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -255,6 +255,9 @@ class NewCommand extends Command
                 default => '/bin/bash -c "$(curl -fsSL https://php.new/install/linux)"',
             };
 
+            $output->writeln('');
+            $output->writeln('  To update, run the following command in your terminal:');
+
             $this->confirmUpdateAndContinue($message, $input, $output);
 
             return;
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -237,8 +237,8 @@ class NewCommand extends Command
 
         if ($isHerd) {
             return $this->confirmUpdateAndContinue(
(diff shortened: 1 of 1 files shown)


src/NewCommand.php
```diff
@@ -199,6 +204,192 @@ protected function ensureExtensionsAreAvailable(InputInterface $input, OutputInt
...
 
+    /**
+     * Check for newer version of the installer package.
+     *
+     * @param  \Symfony\Component\Console\Input\InputInterface  $input
+     * @param  \Symfony\Component\Console\Output\OutputInterface  $output
+     * @return void
+     */
+    protected function checkForUpdate(InputInterface $input, OutputInterface $output)
+    {
```
stancl, line 215: Should this return early if `! $input->isInteractive()`?

src/NewCommand.php
```diff
@@ -199,6 +204,192 @@ protected function ensureExtensionsAreAvailable(InputInterface $input, OutputInt
...
+
+            $this->confirmUpdateAndContinue($message, $input, $output);
+
+            return;
+        }
+
+        if (confirm(label: 'Would you like to update now?')) {
+            $this->runCommands(['composer global update laravel/installer'], $input, $output);
+            $this->proxyLaravelNew($input, $output);
+        }
```
stancl, line 269: There are more ways the installer can be installed besides composer and Herd. Obviously these are the major ones (probably 95%+ of users) but it might be reasonable to somehow check if `laravel` is actually installed via composer.

Not sure how this'd be done on Windows but on Mac/Linux I think this could be as simple as checking that `which laravel` starts with `composer global config home 2>/dev/null`.

PR #450, opened 2025-11-30
Add Node package manager detection

loadinglucian:
Instead of falling back to NPM, detect and prioritize faster package managers (Bun > PNPM > Yarn) over NPM:

-  Added isAvailable() and detect() methods in the NodePackageManager
-  NewCommand tries to detect the best available package manager instead of defaulting to NPM
-  Added tests/NodePackageManagerTest.php to verify the detection and availability logic.


taylorotwell:

PR #461, opened 2026-01-20
Fix current release version

xHeaven:
Updates the CLI version. The project got tagged with `5.24.1`, but the CLI version was not updated accordingly, so it keeps re-prompting for an update.
Master needs a new tag (`5.24.2`) after merging this.


crynobone:
It's better to rely on Composer Runtime Utilities: #462

PR #462, opened 2026-01-20
Utilise Composer 2 runtime utilities

crynobone:
Replace #461


crynobone:
This solution has limitation on Herd, holding off for now

PR #511, opened 2026-05-01
feat: support installing specific Laravel version via --version option

mdMandoinfo:
feat: support installing specific Laravel version via --version option

Adds an optional --version flag to the "new" command, allowing developers
to install a specific Laravel version directly through the installer.

Example:
  laravel new app --version=13.*

Motivation:
Currently, installing a specific Laravel version requires falling back
to "composer create-project", which bypasses the installer's interactive
features such as starter kits and prompts.

This change allows developers to keep the full installer experience
while still selecting a target Laravel version.

Notes:
- Does not change default behavior (latest version is still used)
- Fully backward compatible
- Avoids introducing additional complexity to the install flow

### DIFF
--- a/src/NewCommand.php
+++ b/src/NewCommand.php
@@ -94,6 +94,7 @@ class NewCommand extends Command
             ->setName('new')
             ->setDescription('Create a new Laravel application')
             ->addArgument('name', InputArgument::REQUIRED)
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #541, opened 2026-07-04
fix: restore missing prompts for user in interactive mode

JustChickNugget:
Restore missing prompts for user in interactive mode. The v5.30.0 fully ignores user opinion on some questions while in the interactive mode. So, launching the `laravel new my-app` will start creating the project without asking some of the questions to the user such as testing framework or if user wants to install Laravel Boost.

How v5.30.0 handles it:

Screenshot 2026-07-04 195043

Screenshot 2026-07-04 195113

How this PR handles it:

Screenshot 2026-07-04 195954

Screenshot 2026-07-04 200118

And also added check for `npm` option in package manager conditions block (when setting default value to `npm` option).


taylorotwell:
I no longer want to ask those questions. There were way too many questions in the installation process. If you want to use PHPUnit for example you should now pass `--phpunit` when running the command. Or, run `--no-boost` if you don't want Boost.

JustChickNugget:
> I no longer want to ask those questions. There were way too many questions in the installation process. If you want to use PHPUnit for example you should now pass `--phpunit` when running the command. Or, run `--no-boost` if you don't want Boost.

Got it. But where is this documented? I need to know about things like this in the future. Maybe I missed something?

UPD 1: At least, I only saw `laravel new --help`.

TheNuggitMan:
This sucks. For newbs it's very useful having these questions, to help us learn why we are making certain choices. Surely the flags should be for those who know what they are doing to speed up their new-project installation? (to add boost or phpunit etc)

PR #547, opened 2026-07-10
Add community starter kit option to interactive prompt

alibori:
Hey there :)

I want to purpose this change in the Laravel installer:

Adds a "Community maintained" choice to the interactive starter-kit stack select. When picked, the installer prompts for a Packagist package (vendor/package) or a Git repository URL and feeds it into the existing --using install path, so no changes to the install machinery are needed.

Input is validated as either a vendor/package name or a URL before use.

Thank you for your time!

Axel


taylorotwell:

PR #557, opened 2026-08-11
Adding new command php artisan dev

codingwithrk:
Taylor Otwell introduced new command at LaraconUS.
After new project install at end still displaying `composer run dev`. So replaced with `php artisan dev` command.


taylorotwell:
They are the same thing 👍

PR #559, opened 2026-08-15
Prevent installer self-update loop (#510)

### LINKED ISSUE #510: Laravel new Installer loop after being prompted to install a newer version of the installer
### Laravel Version

13

### PHP Version

8.5.5

### Database Driver & Version

No DB yet. Trying to create a new project.

### Description

```
WARN  A new version of the Laravel installer is available. You have version 5.25.3 installed, the latest version is 5.26.1.

Would you like to update now?
● Yes / ○ No
 ```

When clicking Yes, it makes like it installs somethings then the message comes back again, indefinately.

### Steps To Reproduce

- Try to install a new Laravel project with an outdated version of the installer. 

- After progressing a few steps, it'll prompt you to install a newer versio of the installer. 

- If you click Yes, it'll loop.

### PR DESCRIPTION
**Pull request by wakqasahmed**
Fixes #510

## Root cause

`checkForUpdate()` prompts the user to update, runs `composer global update laravel/installer --with-all-dependencies`, and then unconditionally calls `proxyLaravelNew()`, which re-execs `laravel <same args>` and exits the current process. The freshly re-exec'd process runs `checkForUpdate()` again from scratch.

If the composer update doesn't actually bump the installed version — reported here with dependency constraint conflicts (`composer global why-not laravel/installer …`), and previously also with Herd's own bundled installer intercepting the `composer`/`laravel` binaries — the re-exec'd process detects the exact same outdated version, shows the same prompt, and the cycle repeats indefinitely if the user keeps confirming.

## Fix

Pass a marker env var (`LARAVEL_INSTALLER_UPDATE_ATTEMPTED`) into the re-exec'd process from `proxyLaravelNew()`. `checkForUpdate()` checks for that marker before showing the update prompt again; if present, it prints a one-line warning and continues with the current version instead of prompting/proxying again. This caps the update flow at a single retry regardless of *why* the update didn't take effect (constraint conflict, Herd override, network/permissions issue, etc.), so it doesn't need to special-case any particular failure mode.

## How this differs from the two previously-closed attempts at #471 (PR #493, PR #477)

Both prior PRs took the same approach: spawn a brand-new `composer global show laravel/installer --format=json` process after the update, parse its JSON output, and `version_compare()` against the pre-update version to decide whether to re-proxy.

- Neither PR was actually rejected on the merits of that approach — #493 was closed unmerged after developing a merge conflict that was never resolved, and #477 was closed by its own author ("moving to fork") without further review.
- That approach still has real gaps: it depends on `composer global show`'s JSON schema/exit code being stable across Composer versions, adds a second shell/process invocation and JSON-parsing path that has to be kept in sync with the actual update command, and doesn't address the specific Herd-intercepted-binary case since `composer` may resolve to something else entirely there.
- This PR instead relies on the existing `runCommands()`/`proxyLaravelNew()` re-exec mechanism the code already uses, tagging the specific re-exec chain instead of querying `composer` again. It's a smaller diff, doesn't introduce a new subprocess call, and caps the loop regardless of which underlying cause (constraint conflict, Herd, network) prevented the version from changing.

## Test plan

- Added `test_it_does_not_prompt_to_update_again_once_an_update_was_already_attempted`, which exercises `checkForUpdate()` directly with a stubbed "latest version" response and confirms it stops prompting/proxying once the marker env var is present.
- Ran the full `tests/NewCommandTest.php` suite locally (PHP 8.5 + Composer 2 container): 16 tests, 34 assertions, 1 pre-existing skip (requires Herd/Valet) — all green.

### CODE REVIEW

**Review by wakqasahmed — COMMENTED** [review]
Cold-start review of the fix for #510.

**Verdict: the mechanism is correct and I could not break it — but the test is both fragile and does not cover the half of the fix that actually matters.**

What I verified positively (no change needed):

- **Env propagation through `runCommands()` is real, not silently dropped.** `['laravel '.$input]` is a list array, so `shouldRunAsTask()` is false (`! array_is_list($commands)` fails) and the call lands on `Process::fromShellCommandline($commandline, $workingPath, $env, ...)` at `src/NewCommand.php:1644`. Symfony merges `$env` over the inherited default env, so the child `laravel` process sees the marker via `getenv()`.
- **Windows is fine.** The Windows interactive branch at `src/NewCommand.php:1720` explicitly builds `$envPairs = array_filter(array_merge($_SERVER, $_ENV, $env), ...)` and passes it to `proc_open`, so the marker survives there too.
- **Precedent is followed.** `LARAVEL_INSTALLER_DEFER_HOOKS` (`src/NewCommand.php:690`) and `LARAVEL_INSTALLER_NO_NODE` (`src/NewCommand.php:1393`) already use the exact same `env:` shape, and `tests/NewCommandTest.php:214` already proves this propagation works end to end.
- **Retries are genuinely capped at 1.** The marker check sits *after* the `version_compare()` early return, so a successful update never trips the warning; and an unsuccessful one warns exactly once because gen-2 never reaches `proxyLaravelNew()` again. I could not construct a >1 loop.
- **The claims about #493 and #477 in the PR body check out.** #493 was closed after `crynobone` asked for an unresolved merge conflict to be fixed; #477 was closed by its own author with "Moving to fork." Neither was rejected on approach.

Specific findings inline below.

File: src/NewCommand.php [hunk 1]
```diff
@@ -423,6 +423,16 @@ protected function checkForUpdate(InputInterface $input, OutputInterface $output
...
+        // If we already proxied into this installer once after prompting the
+        // user to update, avoid prompting (and proxying) again indefinitely
+        // when the update did not actually change the installed version.
+        if (getenv('LARAVEL_INSTALLER_UPDATE_ATTEMPTED') !== false) {
+            $output->writeln('');
+            $output->writeln("  <bg=yellow;fg=black> WARN </> The installer could not be updated to version {$latestVersion}. Continuing with the current version...");
+
+            return;
+        }
+
```
wakqasahmed: **1. The marker leaks into every descendant process, not just the re-exec'd installer.**

Once set on the proxied `laravel` process, this env var is inherited by everything that process spawns for its entire lifetime — `composer create-project`, the starter-kit installer hooks (`runInstallerHooks`), npm, git. If any of those ever shell out to `laravel` (a `post-create-project` hook in a starter kit's `composer.json` is a plausible vector), that nested invocation silently suppresses a legitimate update prompt.

Cheap containment: consume the marker rather than just reading it —

```php
if (getenv('LARAVEL_INSTALLER_UPDATE_ATTEMPTED') !== false) {
    putenv('LARAVEL_INSTALLER_UPDATE_ATTEMPTED');
    ...
}
```

With the CLI default `variables_order=GPCS`, `$_ENV` is empty, so Symfony's `getDefaultEnv()` resolves off `getenv()` and the `putenv()` unset does propagate to grandchildren. Not a blocker, but it turns "suppressed for this whole process tree" into "suppressed for exactly the one check it was aimed at", which is what the comment above claims it does.

**2. `!== false` is looser than the intent.** `LARAVEL_INSTALLER_UPDATE_ATTEMPTED=0` or `=''` in the user's shell also suppresses the prompt, since `getenv()` returns `'0'`/`''`, both `!== false`. The name is namespaced enough that accidental collision is unlikely, but matching on the value you actually set (`=== '1'`) costs nothing and makes an exported-empty-var in someone's dotfiles harmless.

**3. The message is a dead end for the user.** "The installer could not be updated to version X. Continuing with the current version..." tells them what failed but not what to do next, and it fires on the Herd/herd-lite path too — where the installer never attempted an update itself, it only asked the user to. #493 and #477 both landed on "You may need to update manually", which is more actionable. Consider something that names the escape hatch, e.g. pointing at `composer global update laravel/installer --with-all-dependencies` (or the Herd update button) and noting they can re-run `laravel new` afterwards.

**4. Behavior change on the Herd path worth calling out explicitly.** `confirmUpdateAndContinue()` now also carries the marker, so a Herd user who answers "I have updated" while Herd's update is still downloading gets one shot only — gen-2 prints "could not be updated" and proceeds to scaffold with the stale installer, with no re-prompt. That is the correct trade against an infinite loop, but it makes finding #3 (an actionable message) more load-bearing than it looks.

src/NewCommand.php
```diff
@@ -499,7 +509,13 @@ protected function confirmUpdateAndContinue(string $message, InputInterface $inp
...
     {
         $output->writeln('');
-        $this->runCommands(['laravel '.$input], $input, $output, workingPath: getcwd());
+        $this->runCommands(
+            ['laravel '.$input],
+            $input,
+            $output,
+            workingPath: getcwd(),
+            env: ['LARAVEL_INSTALLER_UPDATE_ATTEMPTED' => '1'],
+        );
```
wakqasahmed, line 517: Traced and confirmed working — flagging it here only because it is the load-bearing line of the whole fix and it is the one line with **zero test coverage** (see my comment on the test).

`['laravel '.$input]` is a list, so `shouldRunAsTask()` short-circuits on `! array_is_list($commands)` and this never takes the `runCommandsAsTask()` path; it reaches `Process::fromShellCommandline($commandline, $workingPath, $env, null, null)` at line 1644, where Symfony merges `$env` over the inherited environment. On Windows-interactive it reaches `runCommandsInteractivelyOnWindows()` at line 1720, which merges into `$_SERVER`/`$_ENV` before `proc_open`. Both paths propagate correctly.

tests/NewCommandTest.php
```diff
@@ -307,6 +308,45 @@ public function configureWorkflowPhpVersionPublic(string $directory): void
...
+
+        try {
+            $command->checkForUpdatePublic($input, $output);
+        } finally {
+            putenv('LARAVEL_INSTALLER_UPDATE_ATTEMPTED');
+        }
+
+        $this->assertStringContainsString('could not be updated', $output->fetch());
+    }
+
```
wakqasahmed, line 359: The test is real — it stubs `getLatestVersionData()` so `version_compare('1.0.0', '99.0.0') === -1` genuinely falls through to the new branch, and it cleans up `putenv()` in a `finally`. It is not a trivial assertion. Two problems though:

**1. It tests the cheap half and skips the half that can actually regress.** All this proves is "if the env var is set, print a warning" — a single `getenv()` check. The part of the fix that could plausibly break in future refactoring is `proxyLaravelNew()` handing the marker to `runCommands()` and that marker surviving into the child process. That is completely uncovered. The repo already has the exact pattern for testing this: `test_no_node_option_is_passed_to_laravel_installer_hooks` (line 214) asserts propagation by having the child echo `getenv('LARAVEL_INSTALLER_NO_NODE')`. A sibling test asserting that `proxyLaravelNew()` produces a `Process` whose env carries `LARAVEL_INSTALLER_UPDATE_ATTEMPTED=1` would close the real gap. There is also no negative case — nothing asserts that *without* the marker the prompt path is still reached, so a botched condition inversion would pass.

**2. `$this->agent = new Agent;` makes this test environment-dependent, and it fails rather than skips.** `checkForUpdate()` returns immediately when `$this->agent->isActive()` (line 407), which is `AgentDetector::detect()`. Run this suite from a shell where the detector fires (Claude Code, Cursor, and friends set the env vars it looks for) and `$output->fetch()` is empty, so `assertStringContainsString('could not be updated', ...)` hard-fails for reasons unrelated to the change. The other tests using `new Agent` tolerate this because `isActive()` only gates TTY behavior there; here it is an unconditional early return. Stub `isActive()` to `false` on the anonymous class, or otherwise pin the agent state, so the result does not depend on who is running the suite.

wakqasahmed:
Thanks for the thorough review — addressed all 4 findings in a follow-up commit (a5ec1a2):

1. **Marker leak into descendants** — `checkForUpdate()` now calls `putenv('LARAVEL_INSTALLER_UPDATE_ATTEMPTED')` right after reading the marker, clearing it before the re-exec'd process spawns its own children (composer, npm, starter-kit hooks). Verified this actually unsets `getenv()` for subsequent calls in-process via a new test.
2. **Loose check** — tightened `getenv(...) !== false` to `=== '1'`, so an empty/zero value some user happens to have exported no longer accidentally suppresses the prompt.
3. **Dead-end message** — the suppression warning now reads "...You may need to update manually. Continuing with the current version...", matching the wording used by the previously-closed sibling PRs (#493, #477), and phrased generically enough to be accurate on both the direct-composer-update and Herd/Herd-lite paths (where the installer itself never attempted the update).
4. **Test coverage gaps**:
   - Fixed the fragile test by stubbing `Agent::isActive()` to `false` on an anonymous subclass instead of `new Agent`, so it no longer depends on whether the suite happens to run under an agent-detected environment.
   - Added a negative-case test proving the prompt still fires normally when the marker isn't set.
   - Added a test asserting the marker is actually consumed (`putenv` verified via `getenv()` returning `false` afterward).
   - Added a test exercising the real env-propagation path (`runCommands()` → `Process::fromShellCommandline()`), modeled on the existing `LARAVEL_INSTALLER_NO_NODE` pattern at `tests/NewCommandTest.php:214` — asserts the child process actually receives the marker via `getenv()`, plus the negative case where it isn't passed.

Ran the full suite locally (PHP 8.5, Docker) — all new/updated tests pass, along with the rest of `tests/NewCommandTest.php` (the two unrelated failures I saw locally are network/`composer`-binary-dependent scaffolding tests that also fail on `main` in an offline sandbox and are unaffected by this change). Also ran `phpstan analyse` clean.

Issue #561, opened 2026-08-17
package manager flag ignored when creating a new app with starter kit (initial deps install + ci:check script)


iz-ahmad:
### Installer Version

5.31.1

### Description

When creating a new Laravel app with a starter kit and an explicit Node package manager flag (`--bun`/`--pnpm`/`--yarn`), two things go wrong:

**Problem 1**: **The first dependency install still runs with `npm`, ignoring the flag.**(e.g. `laravel new app --react --bun` installs frontend deps via `npm` the first time, even though `bun` was explicitly requested)
**Problem 2**: **The `ci:check` Composer script is never updated** to use the chosen package manager, rather keeps calling `npm run lint:check` / `npm run format:check` / `npm run types:check` regardless of `--bun`/`--pnpm`/`--yarn`. though `setup` and `dev` scripts get updated correctly.

### Root cause

**# 1 (first install still uses npm):**
For a starter kit, `NewCommand::execute()` (`src/NewCommand.php`) runs things in this order:
Step 1: `composer create-project {starter-kit}` (deferring the kit's own post-install hook; L656-L676).
Step 2: `runInstallerHooks()` - runs `php artisan install:features --ansi` (declared in the kit's `composer.json` under `extra.laravel.installer.post-create-project`; L677).
Step 3: *afterward* does the installer call `installNodeDependencies()` → `determinePackageManager()` (correctly reads `--bun`) → `configureComposerScripts()` (rewrites `setup`/`dev` to say `bun`) → runs the actual `bun install`/`bun run build` (L733).

The problem is **step 2**: `install:features` (in each starter kit's `app/Console/Commands/InstallFeaturesCommand.php`) delegates to `Laravel\Chisel\Chisel::in(base_path())->npm()->packageManager()`. Chisel's `Npm::packageManager()` ([`laravel/chisel` - `src/Node/Npm.php#L43-L61`](https://github.com/laravel/chisel/blob/main/src/Node/Npm.php)) auto-detects purely from disk state: a non-npm lock file, or the `npm`/`bun`/etc. token inside the `dev`/`dev:ssr`/`setup` composer scripts, else **defaults to npm**. 
[...]

PR #15, opened 2015-01-05
Add flag for selecting branch

rweas:
If you use the flag --b you can use whatever branch of laravel you like.

i.e. `laravel new mysite --b=develop`

See #13

PR #563, opened 2026-08-17
Configure correct package manager before running starter kit hooks

iz-ahmad:
### Summary

Addressing the issue #561 problem # 1. When creating a new Laravel app with a starter kit and an explicit Node package manager flag (`--bun`/`--pnpm`/`--yarn`), **the first dependency install still runs with `npm`, ignoring the flag.**(that means, `laravel new app --react --bun` installs frontend deps via `npm` the first time, even though `bun` was explicitly requested).

Please look into the issue #561 for more detailed explanation. In short, **root cause** was: starter kits declare a `post-create-project` installer hook that installs node deps itself, via `Laravel\Chisel`. Chisel resolves which package manager to use by inspecting the application on disk: a non-npm lock file, otherwise the `npm`/`bun`/`pnpm`/`yarn` token inside the `dev`/`dev:ssr`/ `setup` composer scripts, otherwise `npm` as the default. At the point `runInstallerHooks()` (`src/NewCommand.php` L677) is run, `composer.json` still says `npm`- because `configureComposerScripts()` isn't called until later, inside `installNodeDependencies()` (L733). The starter kit ships a `package-lock.json` and no other lock file, so `Chisel` always resolves to that default `npm`. The installer's own bun/correct PM-aware step then runs *afterwards*, and the resulting/final `composer.json` looks correct (with correct package manager defined), even though **the real first install already happened under npm**.

### Changes I made
This PR **resolves the correct package manager and writes it into `composer.json` *before* the hooks run**, so `Chisel` gets the user's choice correctly and the first install runs with the requested manager:

```php
if (! $input->getOption('no-node')) { // now at line 677
    [$packageManager] = $this->determinePackageManager($directory, $input);

    $this->configureComposerScripts($packageManager);
}

$hooksProcess = $this->runInstallerHooks($directory, $input, $output); // now installer hooks run after the package manager is configured
```

`installNodeDependencies()` is left untouched and still resolves and applies the same configuration. Both calls are idempotent. The `--no-node` guard applied as the existing code does in L740.

**No breaking change** with this fix. Applications installed without a starter kit have no installer hooks, so the only difference is that the rewrite now happens slightly earlier (**correctly**) in the run.

**Verified** end-to-end with `laravel new app --react --{pnpm,bun,yarn} --no-interaction`. Before the change, the hook installed via `npm` in every case. After this fix, each run now installs with the requested manager and produces the matching lock file.

> One note: with the fix, `--pnpm` now surfaces a small **issue** in `laravel/chisel`, whose `pnpm install` command *lacks* the `--ignore-scripts` flag that **this laravel/installer (NodePackageManager.php) already passes**, causing `pnpm` 11+ to fail on unapproved dependency build scripts. 
>So, it's **not caused** by this change, rather **it was simply unreachable until now, since the hook always ran `npm`**. As now the hook runs with the correct package manager, thus this issue surfaced itself out. Please check `laravel/chisel` [issue #15](https://github.com/laravel/chisel/issues/15) for details on this. I'll add a fix for this also, and open a PR against `laravel/chisel` for it. Hope that'll resolve the issue. (this issue happens only with `--pnpm`; `--bun`/`--yarn` installs are unaffected.


iz-ahmad:
May I please know what led to the decision to close this PR, @taylorotwell? is it because of the issue that came up in `laravel/chisel`, or is there some other factor? 

understanding your reasoning can help me learn from this and contribute more effectively in the future.. thanks.

